PR #2, opened 2019-05-17
Update AWS Out of Cluster Docs and IAM Policies

mda590:
* Updated the steps to follow AWS' Cost and Utilization report docs - these docs were already following a very similar path to what this doc was walking through previously, so it seemed to make sense to reference that instead.
* Changed instructions from enabling AWS generated tags to enabling user-defined cost allocation tags
* Tightened up the IAM policies so they only allow access to the resources associated with the CUR. Also added additional details about why Kubecost requires each of the permissions. This is also the policy I am running in my Kubecost installation and haven't had any issues with it.
* Reordered the IAM policies so there is one specified for accessing cost information and one specified for accessing resource/inventory information. I think you can expand on this once the key flow changes are made to potentially allow access to a different AWS billing account vs. the account where the cluster exists.

**?:** For the `Inventory and AWS Resource Policy`, It doesn't appear any of these are actually being used except for `ec2:Describe*`. I could be missing something though. If they're not, I think you should remove everything except for the ec2 permission and can come back and add additional resource statements to the policy if/when you start pulling data on other resources via the AWS API.


aws-out-of-cluster.md
```diff
@@ -1,170 +1,168 @@
...
+[Instructions for creating an S3 Bucket to be used for Cost and Usage data](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/setting-up-athena.html#create-athena-bucket)
 
-## Step 3: Enable tags for Billing Reports
-[https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/activate-built-in-tags.html](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/activate-built-in-tags.html)
+## Step 2: Create an AWS Cost and Usage Report
+When creating the Cost and Usage Report, configure the report to be delivered to the bucket created in step #1. When following the instructions for this step, choose `Athena` as the data integration option so that reports are created in parquet format.
 
-## Step 4: Export data to Athena
-[https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/setting-up-athena.html#use-athena-cf](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/setting-up-athena.html#use-athena-cf)
+[Instructions for creating a Cost and Utilization Report](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/setting-up-athena.html#create-athena-cur)
```
dwbrown2: Nit: should this be "Cost and Usage Report"?

dwbrown2:
Thank you again for this contribution! This is a ton easier to follow now.

PR #4, opened 2019-07-25
add kubecost namespace to sample config

AjayTripathy:


custom-prom.md
```diff
@@ -16,7 +16,7 @@ __Required Steps__
       scheme: http
       dns_sd_configs:
       - names:
-        - kubecost-cost-analyzer
+        - kubecost-cost-analyzer.<namespace kubecost runs in>
```
dwbrown2: I think this is fine, but would this be better as a comment? Should the `kubecost` namespace be appended by default?

PR #12, opened 2019-11-21
added steps for using custom Prometheus

atze234:
Solve Problems of 
https://github.com/kubecost/cost-model/issues/239


dwbrown2:
Thanks for this PR! Our product actually depends on a number of other metrics being available. Let me review with the team and follow up based on our discussions.

srpomeroy:
Closing issue as we no longer use [PostgreSQL](https://www.postgresql.org/).

PR #17, opened 2020-04-18
Add additional permission for object ACL

AjayTripathy:
Anything we should put here to note it's only necessary for x-account?


dwbrown2:
Would this yaml support an inline comment? Seems like this would probably be good.

```
# only necessary for cross-account access
```

PR #20, opened 2020-05-20
Thanos Updates

robfaraj:
callout retention policy at point of creation, general formatting, added secret note, AWS yaml formatting


long-term-storage.md
```diff
@@ -110,6 +109,8 @@ $ helm install kubecost/cost-analyzer \
 
 Your deployment should now have Thanos enabled!
 
+> Note: if you delete a previous install of kubecost, do this before creating the secret in step 2
```
dwbrown2, line 112: How about "if you *need* to delete an existing non-Enterprise install"... something to give context on why this would be neeed.

srpomeroy:
Replaced by #123

PR #21, opened 2020-06-24
fix typo

caarlos0:


dwbrown2:
Thanks so much for catching, Carlos!

PR #22, opened 2020-06-30
Fix 'Installation options' link

niiku:


dwbrown2:
Thanks for this, @niiku! Turns out the current syntax works on docs.kubecost.com but not GH.

niiku:
Yeah, I just saw that too :)

PR #38, opened 2020-12-01
Add link to necessary relabeling rules.

AjayTripathy:


dwbrown2:
This is the wrong place to put this IMO. We should add a new diagnostic test and verify that cAdvisor has expected labels.

srpomeroy:
@AjayTripathy @dwbrown2 Mind if I close this? Looks like we added the diagnostic check for cAdvisor labels.
image

PR #39, opened 2020-12-04
Clarifications on how to use the IAM role

AjayTripathy:


dwbrown2 (commented):
I find this update to be very confusing. 

`Enable IAM roles.` Is this needed?

`In these steps, you'll create a Role.` Would it be better to say.. "First, create a Role by following these steps"

`When asked to attach policies, you'll want to attach the policies created above (either manually or via cloudformation)` When are you asked to do this? Can you clarify?

`Then, you need to add an annotation to that service account as described in these docs
This annotation can be added to the kubecost service account by setting .Values.serviceAccount.annotations in the helm chart to eks.amazonaws.com/role-arn: arn:aws:iam::<AWS_ACCOUNT_ID>:role/<IAM_ROLE_NAME>`

Are these two separate bullet points closely related? If so, should you combine?

aws-cloud-integrations.md
```diff
@@ -337,8 +337,10 @@ Now that the policies have been created, we will need to attach those policies t
 	<summary>Attach via Pod Annotation on EKS</summary>
 	
 * Enable IAM roles.
```
dwbrown2: looks like this can now go? otherwise, LGTM

PR #40, opened 2020-12-04
Clarify what is overwritten and what relabel rues create new labes.

AjayTripathy:


dwbrown2 (approved):
LGTM, it would also be helpful to add something to diagnostics.md

PR #50, opened 2021-01-13
Create annotations.md

AjayTripathy:


dwbrown2 (approved):
LGTM, but can we also add a section on how these new metrics can be used?

AjayTripathy:
> LGTM, but can we also add a section on how these new metrics can be used?

It's to do, they can't yet be used? just emitted.

PR #52, opened 2021-01-20
Delete postgres support

AjayTripathy:
This is deprecated.


dwbrown2:
Do we reference this in other locations? I believe we do. I propose we just mark as deprecated for now.

michaelmdresser (approved):
Looks like we've got a merge conflict, but if this is still in our documentation we should probably remove it. I doubt anyone has reason to reference this any more.

Also, this change should probably be re-targeted at `main` instead of `master`.

srpomeroy:
This section was removed in a newer PR against the new default `main` branch.

PR #53, opened 2021-01-20
First revision on issues I encountered

Sean-Holcomb:


dwbrown2:
Great improvements! A little surprised people were able to complete without these. @AjayTripathy can you please review?

Sean-Holcomb:
So what was step 2 and now is step 3 is enabling the tags, tags must exist before they are enabled, @AjayTripathy mentioned that users might already have some tags, but I think adding in additional tags before enabling tags makes more sense.

AjayTripathy (approved):

PR #57, opened 2021-01-28
add methodology doc

Sean-Holcomb:


AjayTripathy:
Call the doc ".md" so the markdown formatter can run, then LGTM if that looks fine.

PR #61, opened 2021-02-03
Update federated-clusters.md

AjayTripathy:
clarify instructions when using values-thanos.yaml


dwbrown2:
What does this mean? Can we clarify? `"(if overriding these values from a values-thanos.yaml file supplied from command line)"`

dwbrown2 (approved):

PR #66, opened 2021-03-10
Update aws-cloud-integrations.md

dsterz:
Hint and official docu link not using dashes but only underscores for tables and views in athena


AjayTripathy:
Thank you for this clarification @dsterz ! Much appreciated.

PR #68, opened 2021-03-11
New allocation docs

nikovacevic:


allocation-new.md
```diff
@@ -0,0 +1,301 @@
...
+  -d accumulate=false \
+  -d shareIdle=false
+  -G
+```
+```json
+{
+  "code": 200,
+  "data": [
+    {
+      "__idle__": { ... },
```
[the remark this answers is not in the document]
nikovacevic: GitHub does the red styling. It's just meant to indicate that there is a rich object in there, but the docs would be 2000 lines long if we included it all, making more coherent to elide the content.

allocation-new.md
```diff
@@ -0,0 +1,301 @@
+
+# Allocation
```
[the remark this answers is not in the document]
nikovacevic: I was honestly pretty confused about the current state of affairs, so I originally just made a new page that made sense from the perspective of our `Allocation` types and the `/model/allocation` API. We call multiple other things "allocation" in different contexts, and I didn't want to under-cut anything.

IMO we ultimately need to:
- deprecate `aggregatedCostModel` altogether
- document, but encourage limited use of `costDataModelRange` (or deprecate it, too, but perhaps there are legitimate use-cases for it?)
- have robust, organized, and opinionated documentation that guides users to the correct pages, based on which questions they are trying to answer

allocation-new.md
```diff
@@ -0,0 +1,301 @@
...
+- `__unmounted__` (or "Unmounted PVs") refers to the resources used by PersistentVolumes that aren't mounted to a Pod using a PVC, and thus cannot be allocated to a Pod.
+
+## Querying
+
+{description of ETL: configuration, build, run, etc.}
+
+```
+GET /model/allocation
+```
+Argument | Description
```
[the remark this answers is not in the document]
nikovacevic: Yep, the docs were just WIP. Complete now.

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+- 2021/01/01 00:00:00 until 2021/01/02 00:00:00
+
+See [Querying](#querying) for the full list of arguments and [Examples](#examples) for more example queries.
+
+## Allocation schema (version 1.76)
+Field | Description
+---: | :---
+name | Name of each relevant Kubernetes concept described by the allocation, delimited by slashes, e.g. "cluster/node/namespace/pod/container"
+properties | Map of name-to-value for all relevant property fields, including: `cluster`, `node`, `namespace`, `controller`, `controllerKind`, `pod`, `container`, `labels`, `annotation`, etc.
+window | Period of time over which the allocation is defined.
```
michaelmdresser, line 68: If this is the same as the window of the precomputed 24 hr sets mentioned earlier, mention that in this description because there's also the query parameter `window` with different meaning.

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+- 2021/01/03 00:00:00 until 2021/01/04 00:00:00
+- 2021/01/02 00:00:00 until 2021/01/03 00:00:00
+- 2021/01/01 00:00:00 until 2021/01/02 00:00:00
+
+See [Querying](#querying) for the full list of arguments and [Examples](#examples) for more example queries.
+
+## Allocation schema (version 1.76)
+Field | Description
+---: | :---
+name | Name of each relevant Kubernetes concept described by the allocation, delimited by slashes, e.g. "cluster/node/namespace/pod/container"
```
michaelmdresser, line 66: Maybe note that the schema differs by aggregation or even include the different schemas for each aggregation (though we may not want to commit to them).

allocation.md
michaelmdresser: Actually, this is probably my own repeated confusion about `window`. I assume this is computed based on the `window=3d` query parameter, and then `start` and `end` are what I was talking about in the first comment. Could we then modify this definition to be more clear about what "over which the allocation is defined" means?

Something like: "Period of time (format: `FORMAT`) defined by the time of query and the `window` parameter. This will be the same for every object in `data[]`."

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+window | Period of time over which the allocation is defined.
+start | Precise starting time of the allocation. By definition must be within the window.
+end | Precise ending time of the allocation. By definition must be within the window.
+minutes | Number of minutes running; i.e. the minutes from `start` until `end`.
+cpuCores | Average number of CPU cores allocated while running.
+cpuCoreRequestAverage | Average number of CPU cores requested while running.
+cpuCoreUsageAverage | Average number of CPU cores used while running.
+cpuCoreHours | Cumulative CPU core-hours allocated.
+cpuCost | Cumulative cost of allocated CPU core-hours.
+cpuEfficiency | Fraction of CPU request that was used.
```
michaelmdresser, line 77: Is this "Average fraction of CPU request that was used while running"?

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+networkCost | Cumulative cost of network usage.
+pvBytes | Average number of bytes of PersistentVolumes allocated while running.
+pvByteHours | Cumulative PersistenVolume byte-hours allocated.
+pvCost | Cumulative cost of allocated PersistentVolume byte-hours.
+ramBytes | Average number of RAM bytes allocated while running.
+ramByteRequestAverage | Average number of RAM bytes allocated while running.
+ramByteUsageAverage | Average number of RAM bytes used while running.
+ramByteHours | Cumulative RAM byte-hours allocated.
+ramCost | Cumulative cost of allocated RAM byte-hours.
+ramEfficiency | Fraction of RAM request that was used.
```
michaelmdresser, line 89: Likewise.

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+cpuCoreRequestAverage | Average number of CPU cores requested while running.
+cpuCoreUsageAverage | Average number of CPU cores used while running.
+cpuCoreHours | Cumulative CPU core-hours allocated.
+cpuCost | Cumulative cost of allocated CPU core-hours.
+cpuEfficiency | Fraction of CPU request that was used.
+gpuHours | Cumulative GPU-hours allocated.
+gpuCost | Cumulative cost of allocated GPU-hours.
+networkCost | Cumulative cost of network usage.
+pvBytes | Average number of bytes of PersistentVolumes allocated while running.
+pvByteHours | Cumulative PersistenVolume byte-hours allocated.
```
michaelmdresser, line 82:
```suggestion
pvByteHours | Cumulative PersistentVolume byte-hours allocated.
```

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+ramBytes | Average number of RAM bytes allocated while running.
+ramByteRequestAverage | Average number of RAM bytes allocated while running.
+ramByteUsageAverage | Average number of RAM bytes used while running.
+ramByteHours | Cumulative RAM byte-hours allocated.
+ramCost | Cumulative cost of allocated RAM byte-hours.
+ramEfficiency | Fraction of RAM request that was used.
+sharedCost | Cumulative cost of shared resources, including: shared namespaces, shared labels, shared overhead.
+externalCost | Cumulative cost of external resources.
+totalCost | Total cumulative cost
+totalEfficiency | Cost-weighted average of `cpuEfficiency` and `ramEfficiency`
```
michaelmdresser, line 93: Should we explain what "cost-weighted average" means?

allocation.md
michaelmdresser: I suppose `properties` should be the canonical location for this information.

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+  "ramEfficiency": 0.913571,
+  "sharedCost": 0,
+  "externalCost": 0,
+  "totalCost": 0.051853,
+  "totalEfficiency": 0.053611
+}
+```
+
+### Special types of allocation
+- `__idle__` refers to resources on a cluster that were not dedicated to a Kubernetes object. An idle resource can be shared (proportionally or evenly) with the other allocations from the same cluster. (See the argument `shareIdle`.)
```
michaelmdresser, line 154:
```suggestion
- `__idle__` refers to resources on a cluster that were not dedicated to a Kubernetes object (e.g. unused CPU core-hours on a node). An idle resource can be shared (proportionally or evenly) with the other allocations from the same cluster. (See the argument `shareIdle`.)
```

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+- `__unallocated__` refers to aggregated allocations without the selected `aggregate` field; e.g. aggregating by `label:app` might produce an `__unallocated__` allocation composed of allocation without the `app` label.
```
michaelmdresser, line 155:
```suggestion
- `__unallocated__` refers to aggregated allocations without the selected `aggregate` field; e.g. aggregating by `label:app` might produce an `__unallocated__` allocation composed of allocations without the `app` label.
```

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+- `__unmounted__` (or "Unmounted PVs") refers to the resources used by PersistentVolumes that aren't mounted to a Pod using a PVC, and thus cannot be allocated to a Pod.
+
+## Querying
+
+```
+GET /model/allocation
+```
+Argument | Default | Description
+--: | :--: | :--
+window (required) | — | Duration of time over which to query. Accepts: words like `today`, `week`, `month`, `yesterday`, `lastweek`, `lastmonth`; durations like `30m`, `12h`, `7d`; RFC3339 date pairs like `2021-01-02T15:04:05Z,2021-02-02T15:04:05Z`; unix timestamps like `1578002645,1580681045`.
```
michaelmdresser, line 165:
```suggestion
window (required) | — | Duration of time over which to query. Accepts: words like `today`, `week`, `month`, `yesterday`, `lastweek`, `lastmonth`; durations like `30m`, `12h`, `7d`; comma-separated RFC3339 date pairs like `2021-01-02T15:04:05Z,2021-02-02T15:04:05Z`; comma-separated unix timestamp (seconds) pairs like `1578002645,1580681045`.
```

allocation.md
```diff
@@ -1,53 +1,406 @@
...
+aggregate | | Field by which to aggregate the results. Accepts: `cluster`, `namespace`, `controllerKind`, `controller`, `service`, `label:<name>`, and `annotation:<name>`. Also accepts comma-separated lists for multi-aggregation, like `namespace,label:app`.
+accumulate | false | If `true`, sum the entire range of sets into a single set.
+idle | true | If `true`, include idle cost (i.e. the cost of the un-allocated assets) as its own allocation. (See [special types of allocation](#special-types-of-allocation).)
+external | false | If `true`, include external costs in each allocation.
```
michaelmdresser, line 169:
```suggestion
external | false | If `true`, include [external costs](http://docs.kubecost.com/getting-started#out-of-cluster) in each allocation.
```

Is there a better link for OOC costs explanation than the install instructions?

michaelmdresser (commented):
Looking awesome. I love the detailed and thoroughly explained examples and, of course, that error bounds table.

allocation.md
nikovacevic: "window" is one of the most loaded terms we have, so unfortunately it does get confusing. I like where you're headed with that suggestion, but even that isn't quite right because for "3d" you will get three different sets, the windows of which will sum to the full queried window, but it won't be the same for every object in `data[]`.

michaelmdresser: Oh, shoot. So `window` is actually the duration "precomputed 24 hr set" mentioned earlier, and then `start` and `end` are the actual bookends of the **allocation**, yeah?

michaelmdresser: Keeping this all straight is a bit tricky still :sweat_smile:

nikovacevic: How does this sound:
> Period of time (format: `"start": "2009-01-02T15:04:05Z07:00", "end": "2009-01-02T15:04:05Z07:00"`) within the queried window, over which the allocation is defined. Will be the same for each allocation in any given object within the `data` array.

allocation.md
nikovacevic: Right. As we've talked about, it might even be nice to deprecate name in favor is _just_ properties once the `Properties` struct is rewritten.

nikovacevic: This was probably the hardest one to define, for this reason.

allocation.md
nikovacevic: If we want to use the word "average", it would be most accurate to say: "Ratio of `cpuCoreUsageAverage`-to-`cpuCoreRequestAverage`, meant to represent the fraction of requested resources that were used." Is that better, or too wordy?

allocation.md
nikovacevic: See above

allocation.md
nikovacevic: Goodness

allocation.md
nikovacevic: I didn't want to get too technical w/r/t equations, but we could try to provide examples somewhere. What would you recommend we add here as a description? The equation would be `((cpuEfficiency * cpuCost) + (ramEfficiency * ramCost)) / (cpuCost + ramCost)`, which isn't terribly long.

allocation.md
nikovacevic: Not to my knowledge. Is this where you would send people, @dwbrown2?

allocation.md
michaelmdresser: Yeah, I think that sounds good. :+1:

allocation.md
michaelmdresser: Looks good as-is then. We don't have to commit to a schema if we're going to deprecate it soon-ish (hopefully :smile:).

allocation.md
michaelmdresser: I definitely prefer that rewording.
```suggestion
cpuEfficiency | Ratio of `cpuCoreUsageAverage`-to-`cpuCoreRequestAverage`, meant to represent the fraction of requested resources that were used.
```

allocation.md
michaelmdresser:
```suggestion
ramEfficiency | Ratio of `ramByteUsageAverage`-to-`ramByteRequestAverage`, meant to represent the fraction of requested resources that were used.
```

allocation.md
michaelmdresser: I'm a fan of just including that equation. I think it can only help understanding.
```suggestion
totalEfficiency | Cost-weighted average of `cpuEfficiency` and `ramEfficiency`. In equation form: `((cpuEfficiency * cpuCost) + (ramEfficiency * ramCost)) / (cpuCost + ramCost)`
```

allocation.md
michaelmdresser: Tagging @dwbrown2 again but I don't think we have to wait if he's busy. This link seems reasonable to me.

michaelmdresser (approved):
Lovin' it. A couple of leftover comments but I'd be fine with a merge any time now. :+1:

PR #71, opened 2021-03-15
docs(allocation): filter params might change results

evolkmann:
Related Slack Discussion: https://kubecost.slack.com/archives/CLFV60Y90/p1615823847095700

### DIFF
--- a/allocation-api.md
+++ b/allocation-api.md
@@ -77,7 +77,8 @@ API parameters include the following:
 * `etl` setting this variable to `true` forces a request to be served by the ETL piepline. More info on this feature in the Caching Overview section below. 
 
 <a name="filter-params"></a>  
-Optional filter parameters include the following:   
+Optional filter parameters. Note that these filters might change the calculated costs, for example when splitting the `sharedNamespaces` only
+across a single namespace instead of splitting among all existing namespaces.
 
 Filter | Description
 --------- | -----------


dwbrown2:
@evolkmann I believe there may be a misunderstanding here. Shared namespaces, labels, etc are not filter params. They are intended to change the result and are covered in the section above. Filters are a different field as covered in this section.

Does that help clarify?

evolkmann:
Hi @dwbrown2, it is well possible that I have a misunderstanding, let me try to explain.

I noticed that the results changed also when applying a filter (e.g. `?namespace=demo`), because then all the `sharedNamespaces`, etc. are no longer shared between all available namespaces, but are summed up on the single `demo` namespace.

For example, consider this situation:
|namespace|cost|`sharedNamespaces=red,green`|`sharedNamespaces=red,green&namespace=green`|
|-|-|-|-|
|red|1|-|-|
|green|1|-|-|
|blue|1|2|-|
|yellow|1|2|3|

My expectation was that `yellow` would still be `2` because I thought the `namespace` parameter is just a way of reducing the amount of returned data.

AjayTripathy:
@evolkmann could you share what version you're on? We've had some changes in filtering logic in 1.76.0, could you upgrade if not on it and take a look?

evolkmann:
@AjayTripathy I am on `v1.75.0`. I will try to perform the update in the coming days and take another look at this. Thanks for the info so far.

evolkmann:
@AjayTripathy @dwbrown2 The results are now the same in `v1.76.0`, regardless if the the `namespace` parameter is supplied 👍🏻

Just to understand what has changed, can you point me to a diff? I did not find it (or oversee) in [kubecost/cost-analyzer-helm-chart/](https://github.com/kubecost/cost-analyzer-helm-chart/compare/v1.75.1...v1.76.0) or any of the other repos.

And I guess we can close this PR then. Thanks for taking your time.

dwbrown2:
Hi Enzo, that diff is actually in one of our closed source repos today.
Great to hear we got it solved though! Anything else we can help with right
now?

Webb
Kubecost, Co-founder

ᐧ

evolkmann:
@dwbrown2 ok, I understand 👍🏻 No other issues at the moment 🙂

PR #75, opened 2021-03-30
Add line about GKE resource omission from OOC

Sean-Holcomb:
Doc update for: https://github.com/kubecost/kubecost-cost-model/pull/366

### DIFF
--- a/gcp-out-of-cluster.md
+++ b/gcp-out-of-cluster.md
@@ -38,4 +38,4 @@ Container:  "kubernetes_container":  &lt;container>
 
 To use an alternative or existing label schema for GCP cloud assets, you may supply these in your [values.yaml](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml) under the "kubecostProductConfigs.labelMappingConfigs.\<aggregation\>\_external_label" 
 
-> note: Google generates its own labels for GKE resources "goog-gke-node" and "goog-gke-volume". Values with these labels are excluded from OOC costs because they are already included in in-cluster costs under the kubernetes service. This will exclude GKE resources from other clusters however, to address this it is recommended to have kubecost installed on each cluster where insights into costs are required.
+> Note: Google generates special labels for GKE resources (e.g. "goog-gke-node", "goog-gke-volume"). Values with these labels are excluded from out-of-cluster costs because Kubecost already includes them as in-cluster assets. Thus, to make sure all cloud assets are included, we recommend installing Kubecost on each cluster where insights into costs are required.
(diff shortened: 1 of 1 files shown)


gcp-out-of-cluster.md
```diff
@@ -38,4 +38,4 @@ Container:  "kubernetes_container":  &lt;container>
 
 To use an alternative or existing label schema for GCP cloud assets, you may supply these in your [values.yaml](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml) under the "kubecostProductConfigs.labelMappingConfigs.\<aggregation\>\_external_label" 
 
-> note: Google generates its own labels for GKE resources "goog-gke-node" and "goog-gke-volume". Values with these labels are excluded from OOC costs because they are already included in in-cluster costs under the kubernetes service. This will exclude GKE resources from other clusters however, to address this it is recommended to have kubecost installed on each cluster where insights into costs are required.
+> Note: To help prevent double counting on in-cluster costs from the exported cost data, GKE items with a kubernetes_cluster label that matches the cluster name set on the settings page are ommitted. Whether or not an items is a GKE resourse is determined by tags that Google automatically generates, "goog-gke-node" and "goog-gke-volume".
```
nikovacevic, line 41: So does this mean that in multi-cluster it'll still double-count all the nodes and disks, except the ones associated with the cluster that Kubecost is running on for the given OOC query?

gcp-out-of-cluster.md
```diff
@@ -37,3 +37,5 @@ Container:  "kubernetes_container":  &lt;container>
 </pre>
 
 To use an alternative or existing label schema for GCP cloud assets, you may supply these in your [values.yaml](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml) under the "kubecostProductConfigs.labelMappingConfigs.\<aggregation\>\_external_label" 
+
+> note: Google generates its own labels for GKE resources "goog-gke-node" and "goog-gke-volume". Values with these labels are excluded from OOC costs because they are already included in in-cluster costs under the kubernetes service. This will exclude GKE resources from other clusters however, to address this it is recommended to have kubecost installed on each cluster where insights into costs are required.
```
nikovacevic, line 41: Some of this verbiage was a little confusing to me. How do you feel about this suggestion?

> Note: Google generates special labels for GKE resources (e.g. "goog-gke-node", "goog-gke-volume"). Values with these labels are excluded from out-of-cluster costs because Kubecost already includes them as in-cluster assets. Thus, to make sure all cloud assets are included, we recommend installing Kubecost on each cluster where insights into costs are required.

Sean-Holcomb: Much more concise, I like it

PR #80, opened 2021-04-26
Add documentation for max headroom right-sizing

michaelmdresser:


apis.md
```diff
@@ -28,5 +28,8 @@ Assets API retrieves the backing cost data broken down by individual assets, e.g
 
 __[/savings](https://docs.google.com/document/d/1h_LQuTdwzbzcSonZ49w0lktaaHsulNsIj-O0OZR12fc/edit)__
 
-Exposes a set of APIs for cost optimization insights, e.g. pod-rightsizing and more. 
+Exposes a set of APIs for cost optimization insights.
 
+__[/savings/requestSizing](./api-request-right-sizing.md)
```
dwbrown2, line 33: do you like this in its own separate doc vs having all in one Savings API page?

potentially, related... feels like its an opportunity to move that Google doc to this repo...

api-abandoned-workloads.md
```diff
@@ -0,0 +1,13 @@
...
+```
+http://<kubecost-address>/model/savings/abandonedWorkloads
+```
+
+## Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| `days` | int | Number of historical days over which network traffic should be measured. |
+| `threshold` | int | The threshold of traffic at which a workload is determined abandoned. |
```
michaelmdresser, line 13: @mbolt35 Do you know what the units are here?

apis.md
michaelmdresser: Moved, let me know what you think.

api-abandoned-workloads.md
nikovacevic: bytes/second

api-request-right-sizing.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+```
+A CronJob that spins up 1 pod with 1 container on a cadence faster than it takes the
+job to run.
+3 | ------
+5 |    ------
+7 |        ------
+Correct max for the CronJob: 7+5 = 12
+```
```
nikovacevic, line 56: I don't feel super strongly about it, but I think all of this is implementation detail that doesn't necessarily need to be included here.

nikovacevic (approved):
Seems good to me, but happy to defer to @dwbrown2 on some specifics

api-abandoned-workloads.md
michaelmdresser: Thanks!

PR #86, opened 2021-05-18
update cost-allocation

kbrwn:
remove mentions of annotations as only labels work for cost allocation at this time.


dwbrown2:
LGTM, thanks for updating this!

PR #87, opened 2021-05-21
Add allocation params

Sean-Holcomb:
Add new allocation api parameters to doc

### DIFF
--- a/allocation.md
+++ b/allocation.md
@@ -178,9 +178,9 @@ filterLabels | | Comma-separated list of annotations to match; e.g. `app:cost-an
 filterServices | | Comma-separated list of services to match; e.g. `frontend-one,frontend-two` will return results with either of those two services.
 shareIdle | false | If `true`, idle cost is allocated proportionally across all non-idle allocations, per-resource. That is, idle CPU cost is shared with each non-idle allocation's CPU cost, according to the percentage of the total CPU cost represented.
 splitIdle | false | If `true`, and `shareIdle == false` Idle Allocations are created on a per cluster or per node basis rather than being aggregated into a single "\_idle\_" allocation.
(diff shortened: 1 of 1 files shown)


allocation.md
```diff
@@ -177,6 +177,10 @@ filterAnnotations | | Comma-separated list of annotations to match; e.g. `name:a
 filterLabels | | Comma-separated list of annotations to match; e.g. `app:cost-analyzer, app:prometheus` will return results with either of those two label key-value-pairs.
 filterServices | | Comma-separated list of services to match; e.g. `frontend-one,frontend-two` will return results with either of those two services.
 shareIdle | false | If `true`, idle cost is allocated proportionally across all non-idle allocations, per-resource. That is, idle CPU cost is shared with each non-idle allocation's CPU cost, according to the percentage of the total CPU cost represented.
+splitIdle | false | If `true`, and `shareIdle == false` Idle Allocations are created on a per cluster or per node basis rather than being aggregated into a single "\_idle\_" allocation.
+idleByNode | false | If `true`, idle allocations are created on a per node basis. This only affects results if `splitIdle == true`
```
dwbrown2, line 181: It really feels to me like `splitIdle` and `idleByNode` could be a single flag that just supports a type? It might just be me but I found this approach confusing.

Sean-Holcomb: It definitely could be, splitIdle was actually already a thing that was not connected to the frontend, I think there could be a few other simplifications with all these arguments by using a idle enum of some sort

Sean-Holcomb: Just updated the description, but it slipped my mind that it also affects sharing, I would still agree that is a lot of flags here and there might be a way to combine some into an enum, but I am unsure on how some of them might be being used today

PR #89, opened 2021-05-22
add final backslash to curl command

kirbsauce:
curl command needed final backslash to execute successfully


dwbrown2:
LGTM! cc @nikovacevic

dwbrown2:
Was able to test and confirm this looks good, thank you!!

PR #99, opened 2021-06-25
Saved reports documentation update

nealormsbee:
Updated saved reports documentation not to include custom label aliases in explanation of the `aggregate` property.


dwbrown2:
Would it make sense to highlight that we do support label breakdowns as documented in this resource? https://github.com/kubecost/docs/blob/master/alerts.md#type-recurring-update

AjayTripathy:
Yes,  we do support custom labels and it does work as described in the docs, @SheepGotoHeaven  I'd just link to the alerts.md doc here instead of removing this.

nealormsbee:
Altered this to reflect @nikovacevic's work to actually support fields like "Product", which alias labels. This is now the complete list of breakdowns supported via the UI, and calls out the required label syntax.

It does not make note of the API support for `annotations`, which does not have a UI option equivalent. Happy to add that if desired.

nikovacevic (approved):
LGTM!

PR #105, opened 2021-07-06
Sean/doc updates methodology

Sean-Holcomb:
- Update Docs that have out of date information
- Add methodology section for allocation docs for reconcile and shareTenantCosts


cost-allocation.md
```diff
@@ -2,23 +2,17 @@
 
 The Kubecost Allocation view allows you to quickly see allocated spend across all native Kubernetes concepts, e.g. namespace, k8s label, and service. It also allows for allocating cost to organizational concepts like team, product/project, department, or environment. This document explains the metrics presented and describes how you can control the data displayed in this view.
 
-![Cost allocation view](cost-allocation.png)
+![Cost allocation view](./images/cost-allocation.png)
```
michaelmdresser, line 5: Love it!

cost-allocation.md
```diff
@@ -32,10 +26,36 @@ kubectl get pods -l 'app notin (prometheus, cost-analyzer, ...)' --all-namespace
...
-### 3. Time window  
-The designated time window for measuring costs. Results for 1d, 2d, 7d, and 30d queries are cached by default.
 
-### 4. Filter  
+### Additional Options
+
+![Options](./images/cost-allocation-options.png)
+
+#### Idle Cost  
+Allocating idle costs proportionately distributes slack or idle _cluster costs_ to tenants. Specifically, this applies to resources that are provisioned but not being fully used or requested by a tenant. As an example, if your cluster is only 25% utilized, as measured by the max of resource usage and requests, applying idle costs would proportionately increase the cost of each pod/namespace/deployment by 4x. This feature can be enabled by default in Settings.
```
michaelmdresser, line 35: This makes sense to me, but I wonder if that's just because I'm already very comfortable with the concept and I wonder about less-familiar users.

The Allocation API doc has this explanation for `shareIdle`: "If true, idle cost is allocated proportionally across all non-idle allocations, per-resource. That is, idle CPU cost is shared with each non-idle allocation's CPU cost, according to the percentage of the total CPU cost represented.". I'm not even sure that that is the most intuitive, either. 

Perhaps an example would help here, like:

Consider a cluster with a total of 8 CPU cores and 20 GB of RAM. It has the following pods:
- 1 requesting 0.25 cores and using 0.1 cores (= 0.25 cores allocated) and requesting 3 GB RAM and using 1 GB RAM (= 3 GB allocated)
- 1 requesting 0.5 cores and using 0.5 cores (= 0.5 cores allocated) and requesting 0.5 GB RAM and using 0.5 GB RAM (= 0.5 GB allocated)
- 1 requesting 1 cores and using 1.25 cores (= 1.25 cores allocated) and requesting 1 GB RAM and using 1.5 GB RAM (= 1.5 GB allocated)

Total CPU allocated is 2 cores.
Total RAM allocated is 5 GB.

Thus, idle CPU is 6 cores and idle RAM is 15 GB. Distributing the idle cost distributes the cost of these idle resources to pods proportionally based on their overall calculated cost.

I'm not entirely convinced that this is the right place for this explanation, though...

@nikovacevic do you have any thoughts here? Perhaps this kind of example should go in some canonical place that explains what idle cost is and everything else can refer to that?

michaelmdresser: Ah, I realize now that this was just relocated from elsewhere in the doc. We can totally merge without changing this -- it just caught my eye and got me thinking.

michaelmdresser (approved):
Huge fan of updating core docs. A couple of questions, but overall it looks great!

cost-allocation.md
nikovacevic: I'm not opposed to examples, but I think that it's better to leave examples for a separate section, or else the "main line" of the documentation can become jumbled and unreadable.

I believe there was talk with @kbrwn about a "Cookbook" section of docs, where we could step through a real-life example of: install the product; run some allocation/assets queries; make some reports; do some request sizing. I think this idea would fit really well in that context.

allocation.md
```diff
@@ -307,6 +307,18 @@ $ curl http://localhost:9090/model/allocation \
 }
 ```
 
+### Notes on Asset Cost Distribution Methodology:
```
nikovacevic, line 310: In general, it's fine to leave out "Notes on" in documentation and just title a section with substance. "Asset Cost Distribution" would be a fine title. I'd also float the idea of calling this "Allocation of Asset Costs" or similar, to make the target of the distribution clear; as in, "Where are the assets costs being distributed? They are being allocated to allocations."

nikovacevic (approved):
Lots of good improvements, @Sean-Holcomb!

cost-allocation.md
michaelmdresser: Sounds excellent :+1:

PR #117, opened 2021-07-28
prom-label-warning

kirbsauce:
In response to the ask for issue being publicly documented, as a start.  https://github.com/kubecost/cost-analyzer-helm-chart/issues/102


dwbrown2:
Nice! Only feedback is that this makes it sound a bit more severe than it is... what about this "may" happen or something similar... I've still never seen this actually happen in the wild.

kirbsauce:
Thanks for the feedback, @dwbrown2.  I was actually on the fence with the tone, so thanks for the nudge.  I'll ease up a little bit.

kirbsauce:
@dwbrown2 , updates look good?

dwbrown2:
LGTM

PR #119, opened 2021-08-18
AjayTripathy-deprecate-cadvisor-changes

AjayTripathy:
We now gracefully handle both legacy and new cadvisor metrics, so this section can be removed.

### DIFF
--- a/custom-prom.md
+++ b/custom-prom.md
@@ -52,36 +52,6 @@ Kubecost uses [Prometheus recording rules](https://prometheus.io/docs/prometheus
 
 > Note: Kubecost recording rules were most recently updated in v1.65.0
 
-### Cadvisor metric labels 
-Kubecost uses `container_name` and `pod_name` labels on cadvisor metrics. For clusters running **k8s v1.16+**, the following relabel config creates the expected labels on cadvisor metrics:
-
-```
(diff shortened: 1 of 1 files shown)


custom-prom.md
```diff
@@ -52,36 +52,6 @@ Kubecost uses [Prometheus recording rules](https://prometheus.io/docs/prometheus
...
-    target_label: pod_name
-    regex: (.+)
-    action: replace
-```
-
-Note that this does not override the source label-- it creates a new label called "pod_name" and copies the value of the pod into it.
-
-On recent versions of **Prometheus Operator**, cadvisor `instance` labels do not match internal Kubernetes node names. This causes usage data to not be registered correctly in Kubecost. The solution is to add the following block into your kubelet/cadvisor scrape config.
-
-```
```
dwbrown2, line 74: @AjayTripathy is this no longer relevant?

AjayTripathy: Yup! The need to do this is gone...this is a holdover from waaaay long ago. we now join metrics based purely on a key of cluster/namespace/pod/container

dwbrown2: Thanks for clarifying, LGTM

PR #127, opened 2021-09-10
Include `namespace` in ALB example

elenalape:
If the namespace is not specified, the load balancer may not be able to access the service


elenalape:
Closing as this change appears in #133

PR #128, opened 2021-09-16
Add page explaining gpu allocation process

kaelanspatel:
A brief overview of the GPU allocation process, with various links to related resources.


gpu-allocation.md
```diff
@@ -0,0 +1,17 @@
+## GPU Allocation
+
+Kubecost performs GPU allocation through [container resource limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/), which are parsed for each pod which has a container resource request or limit of type `nvidia.com/gpu`. A pod requesting GPU resources will have a field `spec.containers.resources` as follows:
```
dwbrown2, line 3: Should "container resource limits" be something like the "container resource block"? Based on the description, looks like either requests or limits can be used. 

Also, silly question.. can you today have a pod that uses GPU without a GPU request or limit?

dwbrown2:
@kaelanspatel looks good, could this just become part of the k8s cost spec?

gpu-allocation.md
kaelanspatel: Yeah, it's true that either requests or limits are used. I think what I was thinking there was that you can have a limit without a request, but not vice-versa, but you're right that being clearer is better.

As for a pod using a GPU without a request... no? I think? You can have a pod with an image that needs a GPU to run e.g. a CUDA program, but without that request/limit it's not going to be using the [device plugin ds](https://github.com/NVIDIA/k8s-device-plugin) and will just error + crashloop as if a gpu doesn't exist.

kaelanspatel:
@dwbrown2 I've hopefully made this doc clearer and linked this it under "learn more" in `cost-allocation.md`. Can you clarify what you mean by "could this just become part of the k8s cost spec"?

dwbrown2:
> Can you clarify what you mean by "could this just become part of the k8s cost spec"?

Added you in a comment to this doc yesterday! Maybe we can add here in addition.

michaelmdresser:
Is this docs change still ready to go? Is it waiting on GPU efficiency to get re-merged @kaelanspatel?

Also, I think we should re-target it at `main` instead of `master` because that should be our default branch.

srpomeroy:
@kaelanspatel Are you able to open up a new PR targeting main with these changes so we can get it merged?

PR #129, opened 2021-09-20
Updates to GCP OOC docs the clarify using values.yaml

srpomeroy:


gcp-out-of-cluster.md
```diff
@@ -43,7 +43,20 @@ In Kubecost, navigate to the settings page and click "update" for the "External
...
+* Set `.Values.kubecostProductConfigs.projectID = <GCP Project ID that contains the BigQuery Export>`
+* Set `.Values.kubecostProductConfigs.gcpSecretName = <Name of the Kubernetes secret that contains the compute-viewer-kubecost-key.json file>`
+* Set `.Values.kubecostProductConfigs.bigQueryBillingDataDataset = <DATASET.TABLE_NAME that contains the billing export>`
+
+Create a secret for the GCP service account key
+> Note: When managing the service account key as a Kubernetes secret, the secret must reference the service account key json file, and that file must be named `compute-viewer-kubecost-key.json`.
+
+``` sh
+kubectl create secret generic gcp-secret -n kubecost --from-file=./compute-viewer-kubecost-key.json
+```
```
michaelmdresser, line 59: I think it would be very helpful to put this step above the helm configuration so that it is immediately obvious what `Name of the Kubernetes secret that contains the compute-viewer-kubecost-key.json file` means.

michaelmdresser (approved):
This is AMAZING. Thank you! This confused me when I first started trying to figure out OOC and of course I didn't update the docs after I figured it out.

One suggestion, but this is good to go if you prefer as-is!

gcp-out-of-cluster.md
srpomeroy: Ah. That makes way more sense. Updated. Thanks!

PR #133, opened 2021-09-29
Add page on Amazon Managed Service for Prometheus (AMP)

elenalape:
The PR adds the docs page on how to use Amazon Managed Service for Prometheus with Kubecost ✅ 

CC @kbrwn

### DIFF
--- a/aws-amp-integration.md
+++ b/aws-amp-integration.md
@@ -127,7 +127,7 @@ ___
 
 To verify that the integration is set up, check that the `Prometheus Status` section on Kubecost Settings page does not contain any errors.
 
-<img width="500" alt="Kubecost status no errors" src="https://user-images.githubusercontent.com/22844059/132998278-fd388e9a-8d61-4b8b-ad1c-0e52f17ca251.png">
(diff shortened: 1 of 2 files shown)


dwbrown2:
Looks great, @elenalape!

@mbolt35 think we're ready to drop recording rules as a requirement?!

mbolt35:
> Looks great, @elenalape!
> 
> @mbolt35 think we're ready to drop recording rules as a requirement?!

@dwbrown2 I'm only familiar with the savings recording rules. Cluster Sizing is still using the following:
```
kubecost_savings_cpu_allocation
kubecost_savings_memory_allocation_bytes
```

So we can remove the following rules:
```
kubecost_savings_running_pods
kubecost_savings_container_cpu_usage_seconds
kubecost_savings_container_memory_usage_bytes
kubecost_savings_pod_requests_cpu_cores
kubecost_savings_pod_requests_memory_bytes
```
from the docs.

elenalape:
Thank you both for the feedback! @mbolt35, I've just removed those rules.

PR #135, opened 2021-09-30
Spot cluster sizing documentation

michaelmdresser:


spot-cluster-sizing.md
```diff
@@ -0,0 +1,80 @@
...
+documentation before introducing spot nodes to your cluster.
+
+Kubecost marking a workload as spot ready is not a guarantee. A domain expert should
+always carefully consider the workload before approving it to run on spot nodes.
+
+### How the recommended cluster configuration is determined
+
+Determing `O` and `S` is achieved by first partitioning all workloads on the
+cluster (based on the results of the Checklist) into sets:
+spot-ready workloads `R` and non-spot-ready workloads `N`. Kubecost consults its
```
Sean-Holcomb, line 43: Is there a way to set workloads to `R` or `N` manually, so that the command output is adjusted accordingly, or would the user just hack and slash at the command as they saw fit?

Sean-Holcomb (approved):
Very descriptive, I feel like I have a good idea of what is going on after reading this

spot-cluster-sizing.md
michaelmdresser: This is an excellent question. Right now, `R` and `N` can only come from the spot checklist which is based on the cluster state and the checklist's logic. Users can manually move a workload from `N` to `R` by using the annotation overrides, but this is clumsy as an experimental tool because it actually modifies cluster state. What I would like to implement, if the spot features get traction, is something on the UI to "hypothetically" mark workloads as spot ready (i.e. move them from `N` to `R` for the sake of experiment) and get a live-updated cluster sizing recommendation.

PR #143, opened 2021-11-03
add new flag for cloud configuration

Sean-Holcomb:
Add description of new flag add in
https://github.com/kubecost/cost-analyzer-helm-chart/pull/1131
for 
https://github.com/kubecost/kubecost-cost-model/pull/569


cloud-integration.md
```diff
@@ -22,7 +22,7 @@ Value | Default | Description
 --: | :--: | :--
 `.Values.kubecostModel.etlAssetReconciliationEnabled` | true | Enables Reconciliation processes and endpoints. Corresponds to `ETL_ASSET_RECONCILIATION_ENABLED` environment variable.
 `.Values.kubecostModel.etlCloudAssets` | true | Enables Cloud Assets processes and endpoints. Corresponds to `ETL_CLOUD_ASSETS_ENABLED` environment variable.
-
+`.Values.kubecostModel.cloudAssetsExcludeProviderID` | false | Excludes Provider ID for Cloud Asset process so that cloud assets are agregatted on service and user label. Corresponds to `CLOUD_ASSETS_EXCLUDE_PROVIDER_ID` environment variable. Currently only available on AWS
```
dwbrown2, line 25: @Sean-Holcomb, what would drilldown look like in this scenario? Any other product impact?

Sean-Holcomb: There are some set backs detailed [here](https://github.com/kubecost/kubecost-cost-model/pull/569). The code for this one might change before it goes in so I will leave it as is for now

nikovacevic: @dwbrown2 basically, this would turn AWS into what GCP is today; i.e. service-oriented, not line-item-oriented. We'd need to implement the "clever subtraction" idea to make the numbers square.

michaelmdresser: `s/agregatted/aggregated/`

michaelmdresser: "aggregated on service and user label" instead of what?

Is there a short explanation of why someone would enable this? Does it reduce the size of the Asset ETL/improve query performance/something else?

PR #144, opened 2021-11-04
Update Azure Cost Export configuration details

srpomeroy:
Adds additional details to how to configure the Azure Cost Export.


Sean-Holcomb (approved):
Definitely a good addition considering what you found. Style improvements are also good

srpomeroy:
@Sean-Holcomb I was unsure of what we prefer for partitioning.

Sean-Holcomb:
We do not, I don't believe that was an option when I wrote this, but it may be an opportunity for optimization in the future.

PR #150, opened 2021-11-24
Kubecost Agent/Hosted Install Docs

mbolt35:


michaelmdresser:
Do we want these docs to be public?

AjayTripathy:
I'm ok with making these public since they'll still need the kubecost-agent key and we shouldn't randomly end up with data sent to us :)

PR #151, opened 2021-11-29
Create thanos-s3-encryption.md

AjayTripathy:
A user guide for encrypting thanos data in s3 in a way that keeps costs low.


dwbrown2:
LGTM, but is there an existing thanos doc that this would fit in? Otherwise it feels harder to discover...

AjayTripathy:
Yeah, will add it to the AWS thanos docs.

AjayTripathy:
See https://github.com/kubecost/docs/pull/152

PR #154, opened 2021-12-09
share -> shareByNode / shareByCluster

nealormsbee:
Updated saved report docs to use shareByNode and shareByCluster rather than 'share', which is no longer supported

### DIFF
--- a/saved-reports.md
+++ b/saved-reports.md
@@ -42,7 +42,7 @@ The following fields apply to each map item under the `reports` key:
   * `service`
   * `statefulset`
   * `team`
-* `idle` idle cost allocation, supports `hide`, `share`, and `separate`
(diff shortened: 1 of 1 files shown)


saved-reports.md
```diff
@@ -42,7 +42,7 @@ The following fields apply to each map item under the `reports` key:
   * `service`
   * `statefulset`
   * `team`
-* `idle` idle cost allocation, supports `hide`, `share`, and `separate`
+* `idle` idle cost allocation, supports `hide`, `shareByNode`, `shareByCluster` and `separate`
```
nikovacevic, line 45: Not an Oxford comma guy, I see... :eyes:

nikovacevic (approved):

saved-reports.md
nealormsbee: My Oxford comma stance has oscillated between pro- and con- many times over the years. I can't even remember what I actually believe anymore

PR #157, opened 2022-01-06
remove documentation for setting values.yaml to configure azure storage

Sean-Holcomb:
Remove description of deprecated azure configuration option so that new users will no longer use it.


michaelmdresser (approved):
LGTM!

PR #166, opened 2022-01-31
Update release process to reflect 2022 changes

AjayTripathy:


release-process.md
```diff
@@ -4,7 +4,8 @@ A Kubecost Release is a snapshot of the source, build output, artifacts, and oth
 
 ## Production releases 
 
-* Production releases are scheduled for the 2nd and 4th Tuesday of every month.
+* Release candidates are scheduled for the last Tuesday of every month.
```
dwbrown2, line 7: Is this the right expectation to set for RC builds? I personally think release candidates should have more flexibility on when they are available...

dwbrown2: also, see the section about installing an RC build below... is this still the correct/best way?

dwbrown2 (approved):

release-process.md
AjayTripathy: Yes, that will still work. I'm ok to remove the expectation on rc's though.

PR #179, opened 2022-02-15
spom: update cost allocation image paths to be absolute

srpomeroy:


dwbrown2:
@srpomeroy thanks for submitting! do these relative links break in certain places?

srpomeroy:
> 

Images/artifacts cannot be stored directly in ZD, so their paths need to point to github.

dwbrown2 (approved):

PR #197, opened 2022-03-09
Update image paths and branches to main

srpomeroy:
Update image paths to be absolute and also switch all docs branch references from `master` to `main`.


srpomeroy:
@kbrwn In updating some broken images, I noticed a handful of doc links that pointed back to the `master` branch. I didn't think there would be any harm in updating them to `main`. But you would know better than I would. Let me know if you think I should revert them.

AjayTripathy (approved):

PR #201, opened 2022-03-15
1-click request sizing ("apply") feature guide and API reference

michaelmdresser:
We don't have to merge this until we fully announce the feature, but I wanted to put it out for review so I can move on to other things.


dwbrown2:
Nice, here's initial feedback!

* can we make it easier for users to understand how to enable this feature, i.e. show that it's just one flag. This feels like a relevant example: https://github.com/kubecost/docs/blob/main/network-allocation.md
* also, can we highlight the benefits just a little more in the intro?

michaelmdresser:
Both done. I'll make an even stronger pitch for the feature when the frontend is ready to go -- I'll write a mini tutorial for using the new feature in UI and API, perhaps as a blog post.

dwbrown2:
> Good call. Clarified, let me know what you think.

Oh, I see why you added this comment now. Would be nice to make the service key optional.

michaelmdresser:
Service key will be optional when we merge:
- https://github.com/kubecost/cluster-controller/pull/6
- https://github.com/kubecost/cost-analyzer-helm-chart/pull/1379

dwbrown2:
Reading this doc it feels hard to understand how to enable/use this feature... isn't it just this section that is required? https://github.com/kubecost/docs/blob/8f7f9cbc603f9e57a792e6a3762a4ac4457a401a/controller.md#deploying

Could we simplify? Here's an example that comes to mind:
https://guide.kubecost.com/hc/en-us/articles/4407595973527-Network-Traffic-Cost-Allocation

michaelmdresser:
@dwbrown2 I've now split the documentation here into an API reference and a how-to guide. Let me know what you think!

michaelmdresser:
As a general theory of documentation, I'm trying to roughly adhere to https://diataxis.fr/ -- this PR introduces "Reference" and "How-to guide" documentation. A blog probably falls in the "Tutorials"  category.

jessegoodier (approved):
Looks great

guide-one-click-request-sizing.md
```diff
@@ -0,0 +1,63 @@
+# Guide: 1-click request sizing
```
srpomeroy, line 1: Need some style guide love here.
https://github.com/kubecost/docs/blob/main/markdown.md#titles

michaelmdresser: Great catch, thanks. Updated.

PR #209, opened 2022-03-30
Added clarity around where there tag could be found in the AWS console.

jcharcalla:
Added details on where to find the cost allocation tags in the AWS console.

### DIFF
--- a/aws-out-of-cluster.md
+++ b/aws-out-of-cluster.md
@@ -26,7 +26,7 @@ To allocate external AWS resources to a Kubernetes concept, use the following ta
 
 *\*In the `kubernetes_label_NAME` tag key, the `NAME` portion should appear exactly as the tag appears inside of Kubernetes. For example, for the tag `app.kubernetes.io/name`, this tag key would appear as `kubernetes_label_app.kubernetes.io/name`.*
 
-Additionally users must activate the `eks:cluster-name`
+Additionally users must activate the `eks:cluster-name` tag located in the AWS Managment Console under “AWS Billing > Cost allocation tags”.
 ![EKS cluster name user tag](https://raw.githubusercontent.com/kubecost/docs/main/images/user_eks_cluster_name_tag.png)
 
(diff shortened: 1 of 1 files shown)


aws-out-of-cluster.md
```diff
@@ -26,7 +26,7 @@ To allocate external AWS resources to a Kubernetes concept, use the following ta
 
 *\*In the `kubernetes_label_NAME` tag key, the `NAME` portion should appear exactly as the tag appears inside of Kubernetes. For example, for the tag `app.kubernetes.io/name`, this tag key would appear as `kubernetes_label_app.kubernetes.io/name`.*
 
-Additionally users must activate the `eks:cluster-name`
+Additionally users must activate the `eks:cluster-name` tag located in the AWS Managment Console under “AWS Billing > Cost allocation tags”.
```
dwbrown2, line 29: This line makes it sound lie this tag is required, but it's not clear to me why that's the case. Is this not an optional addition?

jcharcalla: It's also mentioned on the [Cloud Integrations (Line 33)](https://github.com/kubecost/docs/blob/af68e8a4c51657d2ed140fa60a1a6b9506dabc32/cloud-integration.md?plain=1#L33) page.  In that context it sound like it's only required when using the Beta `.Values.kubecostModel.cloudAssetsExcludeProviderID=true` value.

Still unclear to me if this is required for OOC costs, I'll continue to research.

dwbrown2: Good to update? Pretty sure this is optional!

AjayTripathy: I think that might be required for the .Values.kubecostModel.cloudAssetsExcludeProviderID=true but some of that has been in flux. @Sean-Holcomb would know best.

Sean-Holcomb: Sorry the wording is unclear there. It is required for Exclude Provider ID. It is needed for the "clever subtraction" aspect of the feature.

dwbrown2: I'm actually still not sure I know when exactly when it should be used. @Sean-Holcomb you ok to updating the wording or want to take @jcharcalla?

michaelmdresser:
Hey, I don't want this PR to die on the vine! @jcharcalla what did your research turn up?

dwbrown2:
@jcharcalla anything we can help with here?

Sean-Holcomb:
We are no longer using `eks:cluster-name` for the `Exclude_Provider_ID` so I have removed reference to in with a new PR.
https://github.com/kubecost/docs/pull/231

PR #210, opened 2022-03-31
Explain saved reports values vs ui interaction

nealormsbee:
* Modified the Summary section of the saved reports document to explain the interaction between UI-managed and configmap-managed reports
* Removed the footnote on this interaction since it is now explained more fully
* Changed the h3 headings to be h2 for semantic correctness and style reasons

EDIT: Also apparently Prettier changed a lot of `*` to `-` and trimmed trailing whitespace. I'm fine with changing the `-` back to `*` if anyone feels strongly about that

### DIFF
--- a/saved-reports.md
+++ b/saved-reports.md
@@ -4,7 +4,7 @@
 
 Saved reports can be managed via [`values.yaml`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml), as well as via the Kubecost UI. These two methods of managing saved reports can be combined. The following rules apply:
 
-- If Helm chart definitions are _unused_ (`global.savedReports.enabled = false`), reports created via the UI are saved to a persistent volume, if one is configured, and persist across pod restarts.
+- If Helm chart definitions are _unused_ (`global.savedReports.enabled = false`), reports created via the UI are saved to a persistent volume, and persist across pod restarts.
 
 - If Helm chart definitions are _used_ (`global.savedReports.enabled = true`), they are registered on pod start. Reports can still be freely created/deleted via the UI. On pod restart, however, whatever is defined in the configmap will supersede all UI changes.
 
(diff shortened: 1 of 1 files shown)


saved-reports.md
```diff
@@ -1,62 +1,67 @@
-Saved Reports
-=============
+# Saved Reports
 
-### Summary
+## Summary
+
+Saved reports can be managed via [`values.yaml`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml), as well as via the Kubecost UI. These two methods of managing saved reports can be combined. The following rules apply:
+
+- If Helm chart definitions are _unused_ (`global.savedReports.enabled = false`), reports created via the UI are saved to a persistent volume, if one is configured, and persist across pod restarts.
```
dwbrown2, line 7: I think you can drop ", if one is configured," as it feels implied to me.

dwbrown2 (approved):
LGTM, but should we move configmap to UI to it's own section? Think it's fine either way, just wanted to ask the question...

nealormsbee:
>should we move configmap to UI to it's own section?

I tried it in draft and it felt a little choppier, but I can see it working. And it might be read more if it's not in the summary. Maybe a structure like


## Summary

Saved reports can be managed via [`values.yaml`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml), as well as via the Kubecost UI. These two methods of managing saved reports can be combined. This reference outlines the process of configuring saved reports through a values file and provides documentation on the required and optional parameters.

## Saved Report Parameters
...

## Example Helm values.yaml Saved Reports Section
...

## Combining UI Management with `values.yaml`
When defining saved reports via `values.yaml`, (`global.savedReports.enabled = true`), they are registered on pod start. Reports can still be freely created/deleted via the UI. On pod restart, however, whatever is defined in the configmap will supersede all UI changes.

Generally, the configmap, if present, serves as the source of truth at startup.

If saved reports are not provided via `values.yaml` (`global.savedReports.enabled = false`), reports created via the UI are saved to a persistent volume, and persist across pod restarts.

## Troubleshooting
....

----

I'm good with either format. Do you have a preference?

PR #217, opened 2022-04-12
Update long-term-storage.md

kc-adawson:
Updated the long-term-storage doc. Removed the depreciated --name flag, added the new name specification command, and changed "helm install" to "help upgrade --install" to avoid confusion when Kubecost is already installed as a release (if a release by this name doesn't already exist, run an install).


srpomeroy (approved):
Lgtm

PR #221, opened 2022-04-21
Add AWS account-level tags documentation

biancaburtoiu:
This PR adds documentation on how AWS account-level tags are supported in Kubecost Assets view.

addresses https://github.com/kubecost/cost-analyzer-helm-chart/issues/946

### DIFF
--- a/aws-out-of-cluster.md
+++ b/aws-out-of-cluster.md
@@ -50,6 +50,8 @@ If a resource has a label with the same name as an account-level tag, the resour
 
 Modifications incurred on account-level tags may take several hours to update on Kubecost. Note that upon such a modification, historical data going back 15 days will be updated to contain the new tag values.
 
+Your AWS account will need to support the `organizations:ListAccounts` and `organizations:ListTagsForResource` policies to benefit from this feature.
+
 ## Having issues?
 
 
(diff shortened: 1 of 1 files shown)


aws-out-of-cluster.md
```diff
@@ -42,6 +42,16 @@ In order to make the custom Kubecost AWS tags appear on the cost and usage repor
 
 Instructions for enabling user-defined cost allocation tags [here](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/activating-tags.html)
 
+## Viewing account-level tags
+
+You can view AWS account-level tags in Kubecost; tags are applied to all the resources defined under a given AWS account. You can filter AWS resources in the Kubecost Assets View (or API) by account-level tags by adding them ('tag:value') in the Label/Tag filter. 
+
+If a resource has a label with the same name as an account-level tag, the resource label value will take precedence; it won't be overriden by the value of the account-level tag.
```
nikovacevic, line 49: "overriden" => "overridden"

nikovacevic (approved):
Just one typo, but otherwise looks good!

dwbrown2:
Might be worth noting this feature is available in v1.93?

cc @kirbsauce as fyi

kirbsauce:
does the feature also include GCP?

dwbrown2:
Just looking at docs, this change does not appear to. Not sure if Bianca also implemented GCP or not.

biancaburtoiu:
Apologies for the late reply, at the moment this is only valid for AWS! @kirbsauce @dwbrown2

AdamStack18:
@dwbrown2 @kirbsauce Do we need to extend this to work with GCP?

kirbsauce:
I think so, yes. The original request was actually for GCP and AWS was added later...then implemented first.

PR #234, opened 2022-05-09
Update shared tenancy parameter documentation

nikovacevic:


srpomeroy:
@nikovacevic is this still accurate?

dwbrown2:
My understanding is yet. cc @AjayTripathy @michaelmdresser if they can help while Niko is OOO.

nikovacevic:
@srpomeroy Yes, still accurate: https://github.com/kubecost/kubecost-cost-model/blob/develop/pkg/etl/allocation.go#L513-L524

I'll merge now, not sure how I missed this!

PR #236, opened 2022-05-12
adding sample configs from gh repo

jessegoodier:
Edits:
-update ingress to v1 from v1beta1
-added link to other ingress examples in gh repo
-added our gh repo sample config files to cloud integration instructions
-fixed a few spelling errors
-errantly trimmed trailing whitespace from aws-cloud-integrations.md, causing more edits than really just line 8


aws-cloud-integrations.md
```diff
@@ -5,14 +5,15 @@ Kubecost pulls asset prices from the public AWS pricing API by default. To have
 
 Your user will need necessary permissions to create the Cost and Usage Report, add IAM credentials for Athena and S3. Optional permission is the ability to add and execute CloudFormation templates. Note we do not require root access in the AWS account.
 
+A github repository with sample files required in below instructions can be found here: [https://github.com/kubecost/poc-common-configurations/tree/main/aws](https://github.com/kubecost/poc-common-configurations/tree/main/aws)
```
srpomeroy, line 8: Not sure what is tripping me up here. I think it may be referencing sample files but then calling them required. Can we try rewording it to be more specific about what files are available in that repo? I feel like someone may get confused with this statement when we are also referencing CloudFormation Templates and IAM policy docs.

I also think it would be helpful to use the note syntax like below.
Example: `> A GitHub repository...`

jessegoodier: agreed

PR #237, opened 2022-05-12
Windows support documentation

AjayTripathy:
Add notes on the current state of windows support.

### DIFF
--- a/windows.md
+++ b/windows.md
@@ -26,5 +26,5 @@ Collecting data about windows nodes is supported by Kubecost as of v1.93.0; howe
     ```
 
 # Metrics
-  * When tracked by the kubernetes API, nodes should show up with the correct number of pods and resources.
+  * Accurate node and pod data exist by default, since they come from the kubernetes API
   * By default, we will be missing utilization data for pods on window nodes; pods will be billed based on request size. Kubecost can be configured to pick up utilization data for windows nodes; cadvisor must run on these nodes and be scraped.
--- a/windows.md
+++ b/windows.md
@@ -1,8 +1,6 @@
 Windows Node Support
 ======
 
-Collecting data about windows nodes is supported by Kubecost as of v1.93.0; however, we cannot deploy kubecost or data collection daemonsets to windows nodes.
-
 # Deployment
   * The cluster must have at least 1 linux node for the kubecost cost-model to run on
   * Taint the cost-model deployment's pods to run on this node, eg:
@@ -24,7 +22,9 @@ Collecting data about windows nodes is supported by Kubecost as of v1.93.0; howe
                 values: 
                 - linux
     ```
+Today, we cannot deploy Kubecost or data collection daemonsets to windows nodes.
 
 # Metrics
+Collecting data about windows nodes is supported by Kubecost as of v1.93.0.
   * Accurate node and pod data exist by default, since they come from the kubernetes API
-  * By default, we will be missing utilization data for pods on window nodes; pods will be billed based on request size. Kubecost can be configured to pick up utilization data for windows nodes; cadvisor must run on these nodes and be scraped.
+  * By default, we will be missing utilization data for pods on window nodes; pods will be billed based on request size. Kubecost can be configured to pick up utilization data for windows nodes; cadvisor must run on these nodes and be scraped. That scrape will happen by default, however cadvisor running on the node is not necessarily guaranteed, see https://github.com/google/cadvisor/issues/2170
--- a/windows.md
+++ b/windows.md
@@ -1,6 +1,29 @@
 Windows Node Support
 
 Windows nodes are partially supported by kubecost as of v1.93.0. Additional support is coming soon!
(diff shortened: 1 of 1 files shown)


AjayTripathy:
All really great feedback, thanks Jesse!

AjayTripathy:
@jessegoodier PTAL when you can!

windows.md
```diff
@@ -0,0 +1,29 @@
+Windows Node Support
```
dwbrown2, line 1: title formatting? also should we call this "Windows Support" or even generalize to "Supported Operating Systems"?

windows.md
```diff
@@ -0,0 +1,29 @@
...
+
+Windows nodes are partially supported by kubecost as of v1.93.0. Additional support is coming soon!
```
dwbrown2, line 3: capitalize Kubecost

what would you say partial support means? my view is that we either officially support or we don't.

windows.md
```diff
@@ -0,0 +1,29 @@
...
+* Deployment
+  * The cluster must have at least 1 linux node for the kubecost cost-model to run on
```
dwbrown2, line 5: just curious, why do we have this requirement? do we have plans to eliminate?

windows.md
```diff
@@ -0,0 +1,29 @@
...
+          requiredDuringSchedulingIgnoredDuringExecution:  
+            nodeSelectorTerms:  
+            - matchExpressions:  
+              - key: kubernetes.io/os 
+                operator: In 
+                values: 
+                - linux
+    ```
+
+* Metrics
```
dwbrown2, line 26: section head format?

windows.md
```diff
@@ -0,0 +1,29 @@
...
+  * When tracked by the kubernetes API, nodes should show up with the correct number of pods and resources.
+  * By default, we will be missing utilization data for pods on window nodes; pods will be billed based on request size.
```
dwbrown2, line 28: why's this? does cadvisor now collect windows metrics?

windows.md
```diff
@@ -0,0 +1,29 @@
...
+            nodeSelectorTerms:  
+            - matchExpressions:  
+              - key: kubernetes.io/os 
+                operator: In 
+                values: 
+                - linux
+    ```
+
+* Metrics
+  * When tracked by the kubernetes API, nodes should show up with the correct number of pods and resources.
```
dwbrown2, line 27: why's this important/relevant?

windows.md
dwbrown2: should this just be combined with the point below?

windows.md
AjayTripathy: I think this is specifically about windows support since we have deployment instructions below? Formatting updated though.

windows.md
AjayTripathy: Our docker builds for ARM/AMD... it may be possible to build for windows but this would require some testing? https://adamtheautomator.com/docker-build-tag/

windows.md
AjayTripathy: cadvisor can collect some windows metrics...it's just that support is a bit weak: https://github.com/google/cadvisor/issues/2158

windows.md
dwbrown2: Understood, just wondering if users wouldn't benefit from a broader resource, e.g. what OS's does Kubecost support.

windows.md
dwbrown2: It's not clear what you're trying to say... this feels like saying "the product works when you have the Kubernetes API"

Is that what you're saying?

AjayTripathy: Just trying to clarify expectations for what should work and what doesn't work out of the box.

AjayTripathy: Took another pass at the language there.

windows.md
```diff
@@ -0,0 +1,30 @@
+Windows Node Support
+======
+
+Collecting data about windows nodes is supported by Kubecost as of v1.93.0; however, we cannot deploy kubecost or data collection daemonsets to windows nodes.
```
dwbrown2, line 4: > however, we cannot deploy kubecost or data collection daemonsets to windows nodes.
 
Feels like this point can be moved below and replace with what this document does, i.e. gives an overview of how to deploy, limitations, etc.
Let's capitalize Kubecost

windows.md
```diff
@@ -0,0 +1,30 @@
...
+            - matchExpressions:  
+              - key: kubernetes.io/os 
+                operator: In 
+                values: 
+                - linux
+    ```
+
+# Metrics
+  * Accurate node and pod data exist by default, since they come from the kubernetes API
+  * By default, we will be missing utilization data for pods on window nodes; pods will be billed based on request size. Kubecost can be configured to pick up utilization data for windows nodes; cadvisor must run on these nodes and be scraped.
```
dwbrown2, line 30: > Kubecost can be configured to pick up utilization data for windows nodes

How does this work? Is it default?

dwbrown2 (approved):

windows.md
AjayTripathy: K, capitlized and moved below into the two sections.

PR #239, opened 2022-05-17
update description of Exclude Provider ID

Sean-Holcomb:


AjayTripathy:
cc @dwbrown2 is this helpful for large CUR documentation or are you looking for more direct recommendations for when to use the flag vs not during onboarding?

dwbrown2:
Generally looks great to me! A couple minor points/questions but then ready to share with GTM:

> "For insights below the aggregated"

Any more context on what this refers to?

> Drilling down below the aggregate level should be used sparingly

This feels pretty prescriptive to me, i.e. should we just note these requests have a cost? Don't feel strongly but wanted to ask the question.

dwbrown2 (approved):

AdamStack18:
@Sean-Holcomb Are we good to merge this?

PR #243, opened 2022-05-24
Update azure-config.md

amehmoodcapsi:
there was a typo in the command to install helm


AjayTripathy:
Thanks for the fix @amehmoodcapsi !

PR #251, opened 2022-06-01
Updated context switcher documentation.

teevans:
Updated documentation to use context verbiage. 

Definitely want some feedback on the finer points here!

cc: @dwbrown2


multi-cluster.md
```diff
@@ -1,24 +1,25 @@
-Multi-cluster
+Contexts
 =============
 
-Kubecost supports the ability to view cost and health data across multiple Kubernetes clusters and cloud providers.
-Below are the steps for adding an additional cluster on the **Kubecost Business & Enterprise** tier.
+Kubecost supports the ability to monitor multiple **contexts** of clusters. A **context** refers to either a single monitored cluster, or a set of clusters that leverage persistent storage.
```
srpomeroy, line 4: We might want to consider substituting `persistent storage` with `long term storage`, `durable storage`, or some combination of the two. We should also link it to https://github.com/kubecost/docs/blob/main/long-term-storage.md

PR #256, opened 2022-06-03
Update business.md

kc-adawson:
Clarifying the language around Business tier, previous language was vauge and did not differentiate from Enterprise.


srpomeroy (approved):
LGTM

PR #257, opened 2022-06-03
simplify install guide

jessegoodier:
Removed some duplications and pointed to the guides as applicable. 

Updating formatting for ease of use.

Open to feedback.

### DIFF
--- a/custom-prom.md
+++ b/custom-prom.md
@@ -8,7 +8,7 @@ Custom Prometheus and Grafana
 
 When integrating Kubecost with an existing Prometheus, we recommend first installing Kubecost with a bundled Prometheus ([instructions](http://kubecost.com/install)) as a dry run before integrating with an external Prometheus deployment. You can get in touch (support@kubecost.com) or via our [Slack community](https://join.slack.com/t/kubecost/shared_invite/enQtNTA2MjQ1NDUyODE5LWFjYzIzNWE4MDkzMmUyZGU4NjkwMzMyMjIyM2E0NGNmYjExZjBiNjk1YzY5ZDI0ZTNhZDg4NjlkMGRkYzFlZTU) for assistance.
 
-The Kubecost Prometheus deployment is used both as a source and a sink for cost & capacity metrics. It’s optimized to not interfere with other observability instrumentation and by default only contains metrics that are useful to the Kubecost product. This amounts to retaining __70-90% fewer metrics__ than a standard Prometheus deployment.
+The Kubecost Prometheus deployment is used both as a source and a sink for cost & capacity metrics. It’s optimized to not interfere with other observability instrumentation and by default only contains metrics that are useful to the Kubecost product. This results in __70-90% fewer metrics__ than a Prometheus deployment using default settings.
 
 For the best experience, we generally recommend teams use the bundled prometheus-server & grafana but reuse their existing kube-state-metrics and node-exporter deployments if they already exist. This setup allows for the easiest installation process, easiest ongoing maintenance, minimal duplication of metrics, and more flexible metric retention.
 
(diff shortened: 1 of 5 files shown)


dwbrown2:
Nice! Will try to review next week here if that works.

jessegoodier:
So it is clear, I removed Prometheus recording rules from https://github.com/kubecost/docs/blob/main/custom-prom.md#recording-rules
which was obsolete as of 1.90 released in feb 2022.

My thinking is that the docs are just getting so complex with people only adding and not cleaning up.
Please let me know if we should add it back for now.

install.md
```diff
@@ -5,18 +5,18 @@ To get started with OpenCost and Kubecost, **the recommended path is to [install
...
 
-* You can also install directly with the [Kubecost helm chart](https://github.com/kubecost/cost-analyzer-helm-chart/) with helm 3 using the following commands. This provides the same functionality as the step above but doesn't generate a product token for managing tiers or upgrade trials. 
+* You can also install directly with the [Kubecost helm chart](https://github.com/kubecost/cost-analyzer-helm-chart/) with helm 3 using the following commands. This provides the same functionality as the step above but doesn't generate a product token for managing tiers or upgrade trials.
+
+```
+helm repo add kubecost https://kubecost.github.io/cost-analyzer/
+helm upgrade --install kubecost kubecost/cost-analyzer --namespace kubecost --create-namespace
+```
 
-   ```
```
dwbrown2, line 10: Did you want to remove the indentation here or is this a linter adjustment? I personally like it inline with the bullet point, but let me know what you think.

custom-prom.md
```diff
@@ -8,6 +8,10 @@ Custom Prometheus and Grafana
 
 When integrating Kubecost with an existing Prometheus, we recommend first installing Kubecost with a bundled Prometheus ([instructions](http://kubecost.com/install)) as a dry run before integrating with an external Prometheus deployment. You can get in touch (support@kubecost.com) or via our [Slack community](https://join.slack.com/t/kubecost/shared_invite/enQtNTA2MjQ1NDUyODE5LWFjYzIzNWE4MDkzMmUyZGU4NjkwMzMyMjIyM2E0NGNmYjExZjBiNjk1YzY5ZDI0ZTNhZDg4NjlkMGRkYzFlZTU) for assistance.
 
+The Kubecost Prometheus deployment is used both as a source and a sink for cost & capacity metrics. It’s optimized to not interfere with other observability instrumentation and by default only contains metrics that are useful to the Kubecost product. This amounts to retaining __70-90% fewer metrics__ than a standard Prometheus deployment.
```
dwbrown2, line 11: > It’s optimized to not interfere with other observability instrumentation and by default only contains metrics that are useful to the Kubecost product.

Should we say "The default Kubecost Prometheus..." here to clarify the reference?

custom-prom.md
```diff
@@ -8,6 +8,10 @@ Custom Prometheus and Grafana
...
+
+For the best experience, we generally recommend teams use the bundled prometheus-server & grafana but reuse their existing kube-state-metrics and node-exporter deployments if they already exist. This setup allows for the easiest installation process, easiest ongoing maintenance, minimal duplication of metrics, and more flexible metric retention.
```
dwbrown2, line 13: > kube-state-metrics 

This is actually now optional as I understand it. @AjayTripathy can you help clarify. Might be able to remove below as well.

custom-prom.md
```diff
@@ -55,14 +59,6 @@ To confirm this job is successfully scraped by Prometheus, you can view the Targ
 
 <a name="recording-rules"></a>
 
-### Recording Rules
```
dwbrown2, line 58: I think we're ready here, given it's been nearly 5 months

dwbrown2 (commented):
Was able to review custom-prem + install doc! Just a couple comments, but looking really good!

install.md
jessegoodier: With the indentation, the copy was adding spaces at the beginning of the second line when coping the commands. 

Agreed that the formatting looks nicer indented. 

See:
https://guide.kubecost.com/hc/en-us/articles/4407601821207-Installing-Kubecost

The copy works correctly on GH. 
Maybe we should see how this looks on guide and change back if it looks bad?

custom-prom.md
jessegoodier: Maybe this?
"This results in __70-90% fewer metrics__ than a Prometheus deployment using default settings."

custom-prom.md
AjayTripathy: KSM is now optional, that's correct. That being said, we don't heavily test this usecase and having KSM gives us some redundancy for accurate pod times (if kc is down, ksm is up and vice versa). I'd leave it in for now until HA mode at which point we could potentially truly deprecate KSM.

custom-prom.md
AjayTripathy: We're ready!

AjayTripathy (approved):

PR #285, opened 2022-06-29
Update architecture.md

kc-adawson:
Fixing the list of core components KC can run with - This was mistakenly written as 3a/3b, when those don't exist. Correcting this to say 2a (Prom Server) and 2b (KSM)


srpomeroy (approved):
LGTM

jessegoodier (approved):
LGTM

kc-adawson:
Added Node-Exporter (3c) to required list and removed (optional) for now.

PR #293, opened 2022-07-07
Update getting-started storage Math

### LINKED ISSUE #290: Confusion about storage requirements
The storage configuration documentation states the following:

> Where ingested samples can be measured as the average over a recent period, e.g. sum(avg_over_time(scrape_samples_post_metric_relabeling[24h])). On average, Prometheus uses around 1.5-2 bytes per sample. So ingesting 100k samples per minute and retaining for 15 days would demand around 40 GB. It's recommended to add another 20-30% capacity for headroom and WAL. More info on disk sizing [here](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects).

Is my arithmetic wrong, or does the example actually suggest 4GB is needed rather than 40GB?

2bytes per sample * 100k samples per minute * 15 days (21,600 minutes) = 4320000000 bytes = 4.32GB

gz#2123<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/2123'>Zendesk ticket #2123</a>)</i>

keithhand:
A user reported this we were overestimating Prometheus storage in our documents. This fixes that math.

Fixes #290

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -63,7 +63,7 @@ To determine the appropriate disk size, you can use this formula to approximate:
 needed_disk_space = retention_time_minutes * ingested_samples_per_minutes * bytes_per_sample
 ```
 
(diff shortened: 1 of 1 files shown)


nealormsbee (approved):
Looks good to me :+1:

AjayTripathy:
@keithhand this math seems right but let's up the sample count to 1mm and keep the 40gb. Most users are ingesting closer to 1mm samples per minute.

PR #294, opened 2022-07-08
Remove references to Cloud Assets and Replace with Cloud Usage.

Sean-Holcomb:
Updated Cloud Integration documentation with up to date language on Cloud Usage. Update endpoints for Cloud Usage processes

### DIFF
--- a/cloud-integration.md
+++ b/cloud-integration.md
@@ -11,10 +11,10 @@ Integration with the Cloud Service Providers via their respective billing APIs a
 > GCP users should create [detailed billing export](https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables#detailed-usage-cost-data-schema) to gain access to all of Kubecost cloud integration features including [reconciliation](https://github.com/kubecost/docs/blob/main/cloud-integration.md#reconciliation)
 
 ## Cloud Processes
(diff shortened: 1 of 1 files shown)


dwbrown2:
I feel like we're going away from the OpenCost spec with this change. What's the current thinking on usage vs assets?

Sean-Holcomb:
The primary driver of the name change was the removal of Cloud data from assets. Therefor Cloud data it is no longer Asset data and it is confusing to call it such moving forward. The other strong argument for the verbiage is that in-cluster Asset data corresponds to discrete hardware or software products that are being rented. Cloud on the other hand covers the above and a long lists of service usage. So the question is, is Athena Usage an asset? or is Cloud Logging an asset? or Microsoft Operational Insights? To be fair these things are all in Asset tab as it stands, but I think a bottom up approach in distinguishing these ideas, could help to push the product to a more understandable place, especially as advanced reporting improves. If these concepts are muddled at the level of the model, they are likely to be muddled in the product too. In terms of the name itself, I don't love CloudUsage, but it does express the idea accurately, and just "Cloud" is off the table because it is way too general.

srpomeroy (approved):
Changes look straightforward. My only question is, would there be any reason to keep the older API endpoints documented for legacy versions?

Sean-Holcomb:
The new endpoints hit the same handler as the old ones. If people want to use the old ones they can, but I don't think we should advertise that they still exist.

PR #298, opened 2022-07-10
Update the doc with prometheus port number

santoshr1016:
To fix the error while running the below command
```
wget -O- --timeout=2 --tries=1 <your_prometheus_url>/api/v1/status/config
```


bstuder99:
LGTM

bstuder99 (approved):

srpomeroy:
Great suggestion @nealormsbee ! @santoshr1016 are you open to making that suggested change?

In our markdown implementation it would look something like this:
```
> **Note**: In the above example, <your_prometheus_url> may include a port number and/or a custom path name, resulting in a url like, e.g., http://kubecost-prometheus-server.kubecost:9080/prometheus/api/v1/status/config.
```

Issue #19, opened 2020-05-19
Clarification on spot data feed configuration


dvdmuckle:
Hi there, I'm trying to get the spot instance data feed working for a deployment of Kubecost, and I've run into trouble getting it to work fully, part of my confusion lying in how the docs are formatted. For context. this is on a 1.16.0 cluster in AWS, and we are using kiam in order to grant access, rather than access keys and secrets. The IAM role granted to Kubecost is overly permissive in terms of S3, so I don't believe that to be the problem. 

Is the Prometheus metric `node_total_hourly_cost` a trustworthy sign that the spot feed is working? I did notice the metric is available before configuring the spot feed, but after configuring the spot data feed, the metric dropped significantly, which makes me think things might be working.

There's also the issue that I don't see the following logs at all:
```
I1104 00:21:02.905327       1 awsprovider.go:1397] Found spot info {Timestamp:2019-11-03 20:35:48 UTC UsageType:USE2-SpotUsage:t2.micro Operation:RunInstances:SV050 InstanceID:i-05487b228492b1a54 MyBidID:sir-9s4rvgbj MyMaxPrice:0.010 USD MarketPrice:0.004 USD Charge:0.004 USD Version:1}
I1104 00:21:02.922372       1 awsprovider.go:1376] Spot feed version is "#Version: 1.0"
```
However I do see these logs:
```
I0518 17:40:32.750949       1 awsprovider.go:1804] Found 0 spot data files from yesterday
I0518 17:40:32.765787       1 awsprovider.go:1813] Found 2 spot data files from today
```
After crawling through the code a bit, it appears the logs about finding spot data files confirm that Kubecost is able to read from the S3 bucket. The numbers here are also accurate to the files in the bucket itself. Shortly after the above logs, I also see:
```
2020/05/19 03:19:43 error getting addresses: EnvAccessKeyNotFound: AWS_ACCESS_KEY_ID or AWS_ACCESS_KEY not found in environment
2020/05/19 03:19:43 error getting disks: EnvAccessKeyNotFound: AWS_ACCESS_KEY_ID or AWS_ACCESS_KEY not found in environment
```
Which I believe to be generated by the AWS SDK. So, while kiam and the associated role seem to be working, something else looks to be trying to use environment variables that don't exist.

With that in mind, is there something missing in the docs, or something I'm missing in the docs, to get the spot data feed working?

AjayTripathy:
[...]

PR #306, opened 2022-07-26
adding documentation for new alerts cloudReport and assetBudget

avrodrigues5:
This is the documentation once the [PR](https://github.com/kubecost/kubecost-cost-model/pull/873) gets merged

### DIFF
--- a/alerts.md
+++ b/alerts.md
@@ -16,6 +16,10 @@ Kubecost alerts allow teams to receive updates on real-time Kubernetes spend. Th
  
 6. [Cluster Health](#type-cluster-health) -- used to determine if the cluster's health score changes by a specific threshold.
 
+7. [Asset Budget](#type-asset-budget) -- sends an email and/or Slack alert when a particular asset spend crosses a defined threshold.
+
+8. [Cloud Report](#type-cloud-report) -- sends an email and/or Slack alert with asset spend across all or a subset of cloud resources.
+
 Have questions or issues? View our [troubleshooting guide](#troubleshooting).
   
 ## Configuring Alerts in Helm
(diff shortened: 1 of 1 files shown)


alerts.md
```diff
@@ -16,6 +16,10 @@ Kubecost alerts allow teams to receive updates on real-time Kubernetes spend. Th
  
 6. [Cluster Health](#type-cluster-health) -- used to determine if the cluster's health score changes by a specific threshold.
 
+7. [Asset Budget](#type-asset-budget) -- sends an email and/or Slack alert when a particular asset spend crosses a defined threshold.
```
dwbrown2, line 19: How about "when spend for a particular set of assets crosses a defined threshold."

alerts.md
```diff
@@ -165,6 +169,93 @@ Example Helm `values.yaml`:
 
 ----
 
+
+### Type: Asset Budget
+
+Define asset budgets and alert on asset that overruns the threshold set.
```
dwbrown2, line 175: Proposal of something like: 

```
Define asset budgets and alert when cloud or Kubernetes assets overrun the threshold set.
```

alerts.md
```diff
@@ -165,6 +169,93 @@ Example Helm `values.yaml`:
...
+  window: 3d
+  aggregation: cluster
+  filter: ''
+```
+
+----
+
+### Type: Cloud Report
+
+Sends a recurring email and/or Slack alert with a asset summary report of cost and efficiency metrics.
```
dwbrown2, line 215: "an cloud or Kubernetes asset summary report"?

dwbrown2 (commented):
Dropped my comments, @bstuder99 would you be able to take a pass?

alerts.md
```diff
@@ -16,6 +16,10 @@ Kubecost alerts allow teams to receive updates on real-time Kubernetes spend. Th
  
 6. [Cluster Health](#type-cluster-health) -- used to determine if the cluster's health score changes by a specific threshold.
 
+7. [Asset Budget](#type-asset-budget) -- sends an email and/or Slack alert when a particular asset spend crosses a defined threshold.
+
+8. [Cloud Report](#type-cloud-report) -- sends an email and/or Slack alert with asset spend across all or a subset of cloud resources.
+
 Have questions or issues? View our [troubleshooting guide](#troubleshooting).
   
 ## Configuring Alerts in Helm
```
bstuder99, line 25: Can we change all H2 and lower headers to sentence case? "Configuring alerts in Helm", "Asset budget", etc.

alerts.md
```diff
@@ -165,6 +169,93 @@ Example Helm `values.yaml`:
...
+* `category`
+* `cluster`
+* `service`
+* `type`
+* `provider`
+* `account`
+* `providerid`
+* `label` requires the following format: `label:<label_name>=<label_value>`
+
+Example Helm `values.yaml`:
```
bstuder99, line 238: I think values.yaml should be italicized rather than formatted as inline code.

alerts.md
avrodrigues5: Let me know if "Sends a recurring email and/or Slack alert with a cloud and/or Kubernetes assets summary report. " is good.

alerts.md
avrodrigues5: Changed all references of values.yaml to italicize. A Question why do you as a technical writer prefer italicizing over inline in this instance?

alerts.md
avrodrigues5: Don't you think it's better with Heading here? As a reader i can navigate to type-asset-budget when i click on line Number #19. If there's a valid reason to change it to sentences I'm okay with it.

alerts.md
bstuder99: So I may need a bit more context as to what values.yaml is, as it shows up all over the documentation, but if it is referring to a specific file with a name and format, I would italicize it the same way I would with interactable text elements. If we refer vaguely to a .yaml file (or any file format like .jpg, .png, etc), no italics needed. I just don't think we should use '' because I would only use that for code examples.

alerts.md
bstuder99: Yeah keep the headers as they are, I just wouldn't capitalize every word in the headers besides the title of the page.

alerts.md
avrodrigues5: Thanks, Was just asking for future documentation. I think Italicizing makes sense over backticks here. Values.Yaml in simple words is a source of truth for all kubernetes helm deployment.

bstuder99: Okay. If you don't have a link to our internal Style guide I can send that to you.

mbolt35 (approved):
LGTM!

PR #315, opened 2022-08-08
Create opencost-product-comparison.md

bstuder99:
We're moving the OpenCost product comparison table to the Kubecost docs, listed under the General folder. Table has been checked for technical accuracy already.


bstuder99:
@teevans @dwbrown2 I'm sure you're both still trying to catch up while you were out, but can you take five minutes to look this over and review? This issue has been sitting up for a while.

dwbrown2:
Looks really good!! Only feedback that I can see is this use of an * looks a little confusing at first glance. 

image

Think we should we link to this doc from Pricing FAQ?

bstuder99:
That's a great idea, but we're tinkering with the idea of adding an additional column to the Pricing page visual that would include OpenCost as a less technical product comparison. From Mat Wilcox on the current visuals from Anna: "like it but would also like to see opencost as a part of all of them / like the core and then each level adds more functionality" So maybe we should get some people in a call and decide how to approach this?

PR #318, opened 2022-08-17
Fix broken images on guide.kubecost.com

linhlam-kc:
- Update images URL to use direct link instead of relative link to fix broken images issue on guide.kubecost.com.
- Update code block type to display script in better format.


kc-creynold:
lgtm

PR #323, opened 2022-08-23
update arch diagrams and add cloud bill info

srpomeroy:
Images show broken when viewing doc due to the need to use full paths for ZenDesk docs.


jessegoodier (approved):
LGTM

dwbrown2:
Thanks for these changes! @srpomeroy any reason to not keep network-costs marked as optional?

srpomeroy:
@jessegoodier @mbolt35 and I had a conversation last week around the network-costs daemon set. Hoping to make it a default install in the future. Happy to revert and keep it marked as optional.

dwbrown2:
Would be interested in feedback from others on making this default installed. Even if it's default on, though I still think it's ok to call this optional? 

IMO it would be great to have it bundled out of the box by default, but I worry that installing two separate daemonsets is a lot for most users. Not outright opposed, but I'd be more inclined to make it more discoverable within Kubecost UI first.

srpomeroy:
Yep, that makes sense. I'll add optional to network costs. What is the second daemon set?

dwbrown2:
node-exporter

srpomeroy:
new PR to revert and add optional to other components.
https://github.com/kubecost/docs/pull/327

PR #324, opened 2022-08-24
Request right-sizing recommendation API V2

michaelmdresser:
PR introducing API: https://github.com/kubecost/kubecost-cost-model/pull/923


bstuder99:
I've corrected any style changes, LGTM. I'm currently working to overhaul our existing API docs, so we may have to revisit the way any deprecated APIs are handled. The content may end up being removed and placed on a separate internal or external doc.

PR #329, opened 2022-08-30
Update cost-allocation.md

bstuder99:
Making some large updates to the Cost Allocation page. Still a WIP, multiple changes have not been completed

### DIFF
--- a/cost-allocation.md
+++ b/cost-allocation.md
@@ -1,63 +1,75 @@
 Kubernetes Cost Allocation
 ==========================
 
-The Kubecost Allocation view allows you to quickly see allocated spend across all native Kubernetes concepts, e.g. namespace, k8s label, and service. It also allows for allocating cost to organizational concepts like team, product/project, department, or environment. This document explains the metrics presented and describes how you can control the data displayed in this view.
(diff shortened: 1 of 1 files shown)


bstuder99:
Changes are complete including new screenshots, explanations of new UI features, and cleaning up of formatting.

srpomeroy (commented):
This is great @bstuder99!

Two pieces of feedback.
1. "Aggregate filters" is mentioned a few times. I hesitate to call our aggregations filters because they are showing all resources of that aggregation.
2. We support a wild card `*` in the filter section within edit reports. e.g. `namespace=kube*` would return any namespace that begins with `kube`. We should document this.

bstuder99:
1. So when I use the aggregate filter term, it's because that interactive element's text is "Aggregate By". I can reword it when I'm not referring to the UI specifically, is there a better term for this?
2. You're right, I forgot to add that in there. Let me go back and do that.

srpomeroy:
"Aggregate filter" can work for now. I honestly can't think of a better term to use.

srpomeroy (approved):

PR #341, opened 2022-09-07
Update troubleshoot-install.md

bstuder99:


Adam-Stack-PM (approved):
LGTM - Would verify with Kirby or Kate that this is ok to remove.

dwbrown2:
What's the thinking on removing?

bstuder99:
Sure, Webb. The thinking here is that links to Slack and our email seem very arbitrary and inconsistent across the documentation, and its inclusion seems dubious. I had a few thoughts in mind for how to handle this:

- In this specific article, there is a link to Slack and email in the middle of the article (which I wanted to remove) and again at the very bottom of the article. It just feels as though we add this line to articles somewhat generously but also inconsistently.
- I have seen both support@kubecost.com and team@kubecost.com, and nobody seems to know which we want to prioritize for users to contact
- We have LinkedIn and Twitter icons embedded in our guide docs footers. Ross and I are checking to make sure GitBooks can support this, but I think we could add Slack and Email icons to the bottom also.
- Adam raised a good point to me today that some users may only view our docs through Git, in which case frontend add-ons don't apply (just a potential rebuttal to my above point we should consider)
- I think links to contact resources should exist in introductory/setup guides, but not across misc. tutorials
- I'd rather see it on all articles or on none. Maybe it can be added to the article footer where it says "Edit this doc on GitHub"

These are all my considerations for the time being but I'm sure customer service is an incredibly valuable resource and I want to make sure we don't hinder it. I'd love to bring this discussion elsewhere if possible and collect feedback.

dwbrown2:
> It just feels as though we add this line to articles somewhat generously but also inconsistently.

Would you be up for documenting/proposing a consistent framework about when where to offer our support contact info? I have ideas but would love to get more of your input or input from others that have views. 

> We have LinkedIn and Twitter icons embedded in our guide docs footers. Ross and I are checking to make sure GitBooks can support this, but I think we could add Slack and Email icons to the bottom also.

Seems like a good idea to me! 

@kirbsauce what email address do you recommend for docs? Are you looking at doing a community group/mailing list?

kirbsauce:
My initial thought is that we use support@kubecost.com.  It's consistent and trackable in Zendesk.  If we want to use a different email address we can set that up.

bstuder99:
I definitely would agree we use support@ over team@, and start phasing the teams email out of existing docs (I don't think anyone knows why we use it). As for when to use the support@ email, I'm thinking we use it for Setup articles. Once a user as gone through the basic installation docs, they should already be on Slack and likely know how to contact us.

PR #344, opened 2022-09-08
Network description

jcharcalla:
Added additional info on what the network column represents when the network cost daemon set is not deployed.


thomasvn (commented):
Since CUR (Cost and Usage Report) is specific to AWS, perhaps we should reword

"from the CUR Cloud integration"

to

"from the Cloud Service Provider's [billing integration](https://guide.kubecost.com/hc/en-us/articles/4412369153687-Cloud-Integrations)"

Thoughts?

bstuder99 (approved):
I agree with Thomas' suggestion, LGTM.

PR #349, opened 2022-09-09
Simplify OCP Helm install process

linhlam-kc:
Simplify OCP Helm install process:
- Remove git clone steps
- Add few improvements to both deployment method to allow users to just copy and paste to install Kubecost.

cc: @jessegoodier for your visibility. I'm going to make similar changes on Openshift blog post.

### DIFF
--- a/openshift-kubecost-install.md
+++ b/openshift-kubecost-install.md
@@ -48,10 +48,10 @@ helm upgrade --install kubecost \
 
 If you want to install Kubecost with your desired cluster name, you can use the following commands:
 
-> **Note**: remember to replace CLUSTER_NAME by your desired value 
+> **Note**: remember to replace CLUSTER_ID's value by your desired value
 
 ```bash
-export CLUSTER_ID="CLUSTER_NAME"
+export CLUSTER_ID="CLUSTER_OCP"
 
 helm upgrade --install kubecost \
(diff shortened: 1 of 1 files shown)


openshift-kubecost-install.md
```diff
@@ -503,13 +510,20 @@ cortextool rules print \
...
 
 ```Bash
-
-helm upgrade -i -n kubecost kubecost ./cost-analyzer \
-    --set kubecostModel.promClusterIDLabel=cluster \
-    --set global.prometheus.fqdn=https://<REPLACE-WITH-GRAFANA-PROM-QUERY-ENDPOINT> \
-    --set global.prometheus.enabled=false \
-    --set global.prometheus.queryServiceBasicAuthSecretName=dbsecret
-
+export CLUSTER_ID="CLUSTER_OCP"
```
srpomeroy, line 513: I would include the same `Note` you used above about setting the `CLUSTER_OCP` values to something.

srpomeroy (approved):
LGTM. 
One minor proposed change just to be clear on setting cluster id. But not required.

openshift-kubecost-install.md
linhlam-kc: Good catch and good idea. Thank you Sean, I fixed it in the latest commit before merging.

PR #358, opened 2022-09-20
Update saved-reports.md

jcharcalla:
Added missing configuration options and an example.


thomasvn (commented):
Perhaps add `sharedOverhead` and `sharedLabels` in the example `values.yaml` as well?

Besides that, looks great to me!

PR #372, opened 2022-09-30
Azure Cloud integration

praveenkumar0566:
After testing internally we observed the file should be cloud-integration.json but not azure-storage-config.json. Also the format of the content in the json file has been updated to the working version

### DIFF
--- a/azure-out-of-cluster.md
+++ b/azure-out-of-cluster.md
@@ -29,7 +29,8 @@ The values needed to provide access to the Azure Storage Account where cost data
 
 With these values in hand, you can now provide them to Kubecost to allow access to the Azure Storage API.
 
-To create this secret you will need to create a JSON file that **must**  be named cloud-integration.json
+To create this secret you will need to create a JSON file that **MUST** be named cloud-integration.json
+**Note:** The file name should be cloud-integration.json.Any other file name will not work.
 with the following format:
 
 ```
@@ -51,7 +52,7 @@ Once you have the values filled out use this command to create the secret:
 
 `kubectl create secret generic <SECRET_NAME> --from-file=cloud-integration.json -n kubecost`
 
-Once the secret is created, set `.Values.kubecostProductConfigs.azureStorageSecretName` to
+Once the secret is created, set `.Values.kubecostProductConfigs.cloudIntegrationSecret` to
 `<SECRET_NAME>` and upgrade Kubecost via Helm, other values related to Azure Storage (see another method) should not be set.
 
 After a successful configuration of Azure out of cluster costs, upon opening the Assets page of Kubecost costs will be broken down by service and there will no longer be a banner at the top of the screen that says OOC is not configured.
--- a/azure-out-of-cluster.md
+++ b/azure-out-of-cluster.md
@@ -29,23 +29,27 @@ The values needed to provide access to the Azure Storage Account where cost data
 
 With these values in hand, you can now provide them to Kubecost to allow access to the Azure Storage API.
 
-To create this secret you will need to create a JSON file that must be named azure-storage-config.json
+To create this secret you will need to create a JSON file that **must**  be named cloud-integration.json
 with the following format:
 
 ```
 {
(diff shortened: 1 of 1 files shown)


mmurph3 (commented):
LGTM. We also tested this and confirm it works.

srpomeroy:
cloud-integration.json is the new way of doing it as captured in [the following docs.](https://github.com/kubecost/docs/blob/main/multi-cloud.md)

afaik azure-storage-config.json is still a valid way to configure the azure billing integration.
[Here is an existing reference in OpenCost.](https://github.com/opencost/opencost/blob/0094cba41fc0827e1598ceba9f6901739aed6386/pkg/cloud/provider.go#L31)
Used by [this](https://github.com/opencost/opencost/blob/57bda4821a8ac2f91435fd6216d4073006170b89/pkg/cloud/azureprovider.go#L624) function.

Unless something has changed?
cc: @Sean-Holcomb

azure-out-of-cluster.md
```diff
@@ -29,23 +29,27 @@ The values needed to provide access to the Azure Storage Account where cost data
...
+    ]
 }
 ```
 
 Once you have the values filled out use this command to create the secret:
 
-`kubectl create secret generic <SECRET_NAME> --from-file=azure-storage-config.json -n kubecost`
+`kubectl create secret generic <SECRET_NAME> --from-file=cloud-integration.json -n kubecost`
 
 Once the secret is created, set `.Values.kubecostProductConfigs.azureStorageSecretName` to
```
jessegoodier, line 54: I think we can use 
Values.kubecostProductConfigs.cloudIntegrationSecret: "cloud-integration"
for all providers.

jessegoodier:
One quick suggestion, otherwise I think your changes look good.

praveenkumar0566:
Thanks everyone, can we have this PR merged if it looks good? Looks like i don't have merge permission.

azure-out-of-cluster.md
praveenkumar0566: Changed to cloudIntegrationSecre

srpomeroy:
> Thanks everyone, can we have this PR merged if it looks good? Looks like i don't have merge permission.

Per my other comment, I would like the engineering team to comment on whether the azure-storage-config.json is still a valid way to configure before merging.

dwbrown2:
cc @AjayTripathy @Sean-Holcomb

jessegoodier:
> > Thanks everyone, can we have this PR merged if it looks good? Looks like i don't have merge permission.
> 
> Per my other comment, I would like the engineering team to comment on whether the azure-storage-config.json is still a valid way to configure before merging.

Should we just use cloudIntegration unless there's a reason to use the azure specific key?

srpomeroy:
> Should we just use cloudIntegration unless there's a reason to use the azure specific key?
IMO we should be pushing people to use cloud integration for sure. But that doesn't change the fact that this other method may still be supported.

thomasvn:
For your awareness @srpomeroy @jessegoodier 

Although the old method of `.Values.kubecostProductConfig.azureStorageSecretName` may still be supported, it's been shown to no longer function properly in this in this [Github Issue](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1430) and this [internal doc](https://docs.google.com/document/d/1WaUJsmgsyqwSP82RlhETIKmNyaBAvP2PgbVzFJU8BXQ)

praveenkumar0566:
Can someone approve the PR, I need to send the doc to the user with the latest changes.

srpomeroy:
@praveenkumar0566 If your customer needs something immediately, [you can send them the following doc](https://guide.kubecost.com/hc/en-us/articles/4407595968919#azure-1-1) which already covers the Azure cloud integration config. Alternatively, [send them the rendered commit](https://github.com/kubecost/docs/blob/cff22b4be826e4c329cb95fd44a4a72809d847fc/azure-out-of-cluster.md).

I sent a DM to @Sean-Holcomb asking if he could review this PR and conversation.

PR #377, opened 2022-10-03
Update cloud-integration.md

praveenkumar0566:
Added a note to use Amortized costs while setting the cost-management-billing in Azure. There is a user dielma to slect Acutal costs or Amortized costs, after dicussing internally Amortized costs is a better fit for what we are doing.


jessegoodier (approved):
LGTM

PR #386, opened 2022-10-11
Delete partner-metrics.md

bstuder99:
We currently have a Cost Allocation WebUI and a Cost Allocation APIs page, I think the content from this page is not really necessary. It mentions our Kubecost Partners project, but I don't really think this needs to go in our tech docs when the website covers it already.


srpomeroy:
Duplicate of #384

PR #391, opened 2022-10-14
Documentation on where Kubecost deployments are supported

thomasvn:


srpomeroy (approved):
LGTM. The only nit is that we may want to consider making the GitHub links permalinks as they reference specific lines.

thomasvn:
Replaced existing links with permalinks!

env.md
```diff
@@ -1,11 +1,31 @@
...
+* ​Managed Kubernetes clusters (e.g. EKS, GKE, AKS) *most common*
+* ​Kubernetes distributions (e.g. OpenShift, DigitalOcean, Rancher, Tanzu)
+* ​Bootstrapped Kubernetes cluster​
+* On-prem and air-gapped using custom [pricing sheets](https://guide.kubecost.com/hc/en-us/articles/4407601795863#q-how-do-i-configure-prices-for-my-on-premise-assets)
+
+**This list is certainly not exhaustive!** This is simply a list of observations as to where our users run Kubecost based on their questions/feedback.
+
+## Supported Kubernetes Versions
 
 * Kubecost runs in any Kubernetes v1.8 (Aug 2017) or greater environment.
```
dwbrown2, line 27: Side note: are we ready bump this soon? This is really old these days...

thomasvn: I agree that v1.8 is pretty dated ...

However! Explicitly stating which version Kubecost can run on may get ahead of any user questions!

srpomeroy: Are we confident it still runs on older versions? Do we want to support it on older versions?

thomasvn: I personally, am only moderately confident because I haven't run these backwards compatibility tests myself.

Do we know of a team running these backward compatibility tests? If not, which Kubernetes versions are we comfortable saying that we support?

AjayTripathy: For ent. users definitely want to support all the way back. I'm sure it will run, though I'm guessing not all features will be available.

PR #396, opened 2022-10-19
Delete markdown.md

bstuder99:


srpomeroy:
This doc has some important nuances with regard to ZenDesk. We should leave it until we've fully migrated.

bstuder99:
Ok, let me hold this PR for now and we can revisit it.

srpomeroy (dismissed):

PR #413, opened 2022-10-21
Troubleshoot installation issue with PodSecurityPolicy on k8s clusters >= v1.25

thomasvn:
Primary concern is that PodSecurityPolicy has been [removed from Kubernetes v1.25](https://kubernetes.io/docs/concepts/security/pod-security-policy/), however our Helm Chart still defaults to having it enabled for Grafana. I'm going to create a PR to cost-analyzer-helm chart so that we default to [`pspEnabled: false`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/8fd5502925c28c56af38b0c4e66c4ec746761d50/cost-analyzer/charts/grafana/values.yaml#L3).

Despite that fact that I'll be fixing this issue soon, I believe this doc may still be useful to those who use our current/older Helm Charts on newer clusters.

Of course, let me know if anybody feels otherwise!

### DIFF
--- a/troubleshoot-install.md
+++ b/troubleshoot-install.md
@@ -156,6 +156,36 @@ If this is true, you are likely to be hitting a CoreDNS routing issue. We recomm
 1. Go to <https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L13>
 2. Replace ```{{ $serviceName }}.{{ .Release.Namespace }}``` with ```localhost```
 
+## Issue: PodSecurityPolicy CRD is missing for `kubecost-grafana` and `kubecost-cost-analyzer-psp`
+
+PodSecurityPolicy has been [removed from Kubernetes v1.25](https://kubernetes.io/docs/concepts/security/pod-security-policy/). This will result in the following error during install.
+
+```bash
+$ helm install kubecost kubecost/cost-analyzer
+Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [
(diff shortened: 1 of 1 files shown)


troubleshoot-install.md
```diff
@@ -156,6 +156,36 @@ If this is true, you are likely to be hitting a CoreDNS routing issue. We recomm
...
+# Clone the Helm chart
+$ git clone git@github.com:kubecost/cost-analyzer-helm-chart.git
+​
+# Update Grafana's configuration for PodSecurityPolicy
+# 
+# FILE:
+# ./cost-analyzer-helm-chart/cost-analyzer/charts/grafana/values.yaml
+# 
+# CONFIG:
+# .Values.rbac.pspEnabled: false
```
srpomeroy, line 183: Is there a benefit to cloning the chart locally and manually updating the value in the grafana values file as opposed to just setting it directly in our chart via `.Values.grafana.rbac.pspEnabled`?

thomasvn: Thank you for the catch!! I didn't realize you could set a subchart's values by doing this. Learned something new 😊. Fixed!

srpomeroy:
@thomasvn The section is titled "## Issue: PodSecurityPolicy CRD is missing for `kubecost-grafana` and `kubecost-cost-analyzer-psp`" but there are no details on disabling PSPs for cost-analyzer. Do you think we should add the reference for ours as well (`podSecurityPolicy.enabled=false`) as it's enabled by default?

thomasvn:
@srpomeroy Again, thank you for the really great catch. I've made the adjustment!

thomasvn:
@kubecost/documentation-team. When you get the chance, would love another review!

thomasvn:
I've also updated the doc to disable PSP in `networkCosts` and `prometheus` as well. Used the [poc-common-configurations](https://github.com/kubecost/poc-common-configurations/tree/main/psp-disable) as reference.

bstuder99 (approved):
LGTM

PR #416, opened 2022-10-25
Update agent.md

bstuder99:
Clarified the language to refer to Kubecost Cloud and link to our most recent announcement at the start of the article. Would also like to know if the bottom section is left readable by mistake, or should be hidden:

> /# Not generally needed: kubectl logs --namespace kubecost-agent -l app=kubecost-network-costs --prefix=true --all-containers --tail=-1 >>./kubecost_agent_logs.log
> /# logs are typically small, can just send plain text. otherwise: tar -czvf kubecost_agent_logs.tgz ./kubecost_agent_logs.log && rm ./kubecost_agent_logs.log

### DIFF
--- a/agent.md
+++ b/agent.md
@@ -1,9 +1,11 @@
-Installing Agent for Hosted Kubecost (Alpha)
+Installing Agent for Kubecost Cloud
 ============================================
 
 The _kubecost-agent_ is a lightweight Kubecost exporter that sends metrics to hosted kubecost. In order to install the _kubecost-agent_, you will need a specific key provided by Kubecost.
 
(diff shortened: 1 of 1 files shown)


bstuder99:
We should also consider if this page should still exist, or if there's any content here we feel is irrelevant or misleading as of this month.

jessegoodier (approved):

rossfisherkc (commented):
LGTM

srpomeroy:
@jessegoodier @mbolt35 @biancaburtoiu  does the way we do agent for hosted/legacy apply to cloud as well?

Issue #415, opened 2022-10-21
Auditing page is now outdated, no longer supported by WebUI


bstuder99:
I've been updating all articles with WebUI screenshots, and I noticed I cannot recreate these screenshots for an Auditing menu depicted in the article: https://guide.kubecost.com/hc/en-us/articles/4407595925271-Auditing. It looks like this is now outdated, and so the plan should be to either rewrite this for accuracy, or remove it entirely. Thoughts?

dwbrown2:
We've effectively moved the auditing experience into Allocation now that you can drilldown and view rates/costs for each individual container in your infra.

bstuder99:
Ok. Niko was in favor of just scrubbing the article entirely in a separate Slack thread

PR #417, opened 2022-10-25
Delete audit.md

bstuder99:
Based on https://github.com/kubecost/docs/issues/415, this content should be removed from the docs.


rossfisherkc (approved):
Agreed, auditing has been depreciated and can be removed

PR #419, opened 2022-10-25
Added some additional Athena trouble shooting steps

jcharcalla:
Added common Athena trouble shooting steps.

### DIFF
--- a/aws-cloud-integrations.md
+++ b/aws-cloud-integrations.md
@@ -442,7 +442,7 @@ You can also check query history to see if any queries are failing:
 	
 <img width="1792" alt="Screen Shot 2020-12-06 at 9 43 50 PM" src="https://user-images.githubusercontent.com/453512/101319633-24ef5500-3817-11eb-9f87-55a903428936.png">
 
-### Common Athean errors.
+### Common Athena errors.
 
 #### Incorrect bucket in IAM Policy.
 
--- a/aws-cloud-integrations.md
+++ b/aws-cloud-integrations.md
@@ -442,6 +442,39 @@ You can also check query history to see if any queries are failing:
 	
 <img width="1792" alt="Screen Shot 2020-12-06 at 9 43 50 PM" src="https://user-images.githubusercontent.com/453512/101319633-24ef5500-3817-11eb-9f87-55a903428936.png">
 
+### Common Athean errors.
+
+#### Incorrect bucket in IAM Policy.
+
+* **Symptom:** 
+   A similar error to this will be shown on the diagnostics page under "Pricing Sources" on the "Diagnostics" page. You can search for the <Athena Query ID> in the Athena dashboard to find additional info about the error.
+   ```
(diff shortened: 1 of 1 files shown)


aws-cloud-integrations.md
```diff
@@ -442,6 +442,63 @@ You can also check query history to see if any queries are failing:
 	
 <img width="1792" alt="Screen Shot 2020-12-06 at 9 43 50 PM" src="https://user-images.githubusercontent.com/453512/101319633-24ef5500-3817-11eb-9f87-55a903428936.png">
 
+### Common Athena errors.
+
+#### Incorrect bucket in IAM Policy.
+
+* **Symptom:** 
+   A similar error to this will be shown on the diagnostics page under "Pricing Sources" on the "Diagnostics" page. You can search for the <Athena Query ID> in the Athena "Recent queries" dashboard to find additional info about the error.
```
thomasvn, line 450: I'm unsure how "\<Athena Query ID\>" will render in Markdown. You may need to use the escape character "\\" in front of the carrots. Or you can use the code blocks "`".

aws-cloud-integrations.md
```diff
@@ -442,6 +442,63 @@ You can also check query history to see if any queries are failing:
...
+	```
+#### Query not supported
+* **Symptom:** 
+A similar error to this will be shown on the diagnostics page under "Pricing Sources" on the "Diagnostics" page.
+```
+QueryAthenaPaginated: start query error: operation error Athena: StartQueryExecution, https response error StatusCode: 400, RequestID: <Athena Query ID>, InvalidRequestException: Queries of this type are not supported
+```
+
+* **Resolution:**
+While rare, this issue was caused by and Athena instance which failed to provision properly on AWS. The solution was to delete the Athena DB and deploy a new one. To verify this is needed, find the failed query ID in the Athena "Recent queries" dashboard and attempt to manually run the query.
```
thomasvn, line 485: "an Athena instance" **

aws-cloud-integrations.md
```diff
@@ -442,6 +442,63 @@ You can also check query history to see if any queries are failing:
...
+```
+
+* **Resolution:**
+Previously, if you ran a query without specifying a value for Query result location, and the query result location setting was not overridden by a workgroup, Athena created a default location for you.
+Now, before you can run an Athena query in a region in which your account hasn't used Athena previously, you must specify a query result location, or use a workgroup that overrides the query result location setting. While Athena no longer creates a default query results location for you, previously created default aws-athena-query-results-MyAcctID-MyRegion locations remain valid and you can continue to use them.
+https://docs.aws.amazon.com/athena/latest/ug/querying.html#query-results-specify-location
+The bucket should be in the format of:
+`aws-athena-query-results-MyAcctID-MyRegion`
+It may also be required to remove and reinstall Kubecost. If doing this please remeber to backup ETL files prior or contact support for additional assistance.
```
thomasvn, line 500: "remember" **

PR #424, opened 2022-10-27
Update resource-consumption.md

bstuder99:
Added new section for https://github.com/kubecost/cost-analyzer-helm-chart/pull/1752. Would like to elaborate on default values or Helm values if needed similar to the rest of the doc if needed.


mbolt35 (approved):
Looks good! Thanks @bstuder99

Issue #432, opened 2022-11-03
Creation of a Federated ETL doc


bstuder99:
I have been discussing with Kaelan and Adam the creation of a Federated ETL doc using content from a Google Doc linked in the release notes. I'm finding the technical terminology described a bit dense and confusing, so I'm hoping I can get some feedback on my comments to clear things up before the doc is published. I will link the PR here when I have the official draft.

Doc: https://docs.google.com/document/d/1Y0ttRo_YRs80UH-exnufHl32168cTETJA5Ld3ZCOwlc/edit#heading=h.m6e340wq0pvj

kaelanspatel:
@bstuder99 Thanks for taking a look Brett. I've edited the docs and responded to your feedback. I know some of the terminology (Federated vs Federator) is confusing, hopefully its a little more clear now.

bstuder99:
Great, thanks for the feedback. Let me start drafting up a doc

bstuder99:
Doc is live https://guide.kubecost.com/hc/en-us/articles/10015110616599-Federated-ETL

PR #434, opened 2022-11-03
Create federated-etl.md

bstuder99:
Creates a doc in response to https://github.com/kubecost/docs/issues/432

Doc is still being drafted, WIP, no approval is currently needed.

### DIFF
--- a/federated-etl.md
+++ b/federated-etl.md
@@ -31,37 +31,40 @@ This diagram shows an example setup of the Federated ETL with:
 
 The result is 5 clusters federated together.
 
-**insert diagram**
+![Federated ETL diagram](https://raw.githubusercontent.com/kubecost/docs/main/images/federated-etl.png)
 
 ## Setup
-### Step 1: Storage Configuration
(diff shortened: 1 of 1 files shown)


bstuder99:
Doc is 95% complete, the last thing I need to finalize is the capitalization of the misc. vocab. When do we capitalize the different cluster types, storage types, and use of the word federation? I believe most of these terms should be lowercase, but it looks like many still capitalize Federation in the context of Kubernetes.

kaelanspatel:
@bstuder99 Made a quick change there to the diagram to make it better/more accurate. Looks really good!

I would actually say either we keep the capitalization of every concept in the Overview section (ik I wasn't super consistent in the original doc), or we don't capitalize anything. These aren't universal concepts, but things we've created inside Kubecost so either way is ok, whatever we decide. I'll defer to you.

kaelanspatel (approved):

bstuder99:
Ok, I've added some consistency to how terms are capitalized. I'll get the page published and categorized.

PR #435, opened 2022-11-04
Edit style of etl-backup.md

thomasvn:
This is an extension of my previous PR #433 

I realized that Heading4 didn't look nice when Zendesk rendered the markdown. Changing all H4 to bolded text instead.


bstuder99:
This is strange, I'm looking at the frontend and it should usually display H4 as bolded text. LGTM.

bstuder99 (approved):

bstuder99:
It looks like Gitbook displays H4 as bolded text, but I don't think we'll have to revert any of this at a later point.

PR #436, opened 2022-11-04
ContainerStats pipeline documentation

michaelmdresser:
Basic docs for the ContainerStats pipeline so the team (and users) have somewhere to go first if something looks awry.

### DIFF
--- a/containerstats-pipeline.md
+++ b/containerstats-pipeline.md
@@ -1,0 +1,47 @@
+# The ContainerStats pipeline
+
+The ContainerStats pipeline builds statistical representations of individual
+containers' resource usage over time.  The pipeline is part of the `cost-model` container.
+
+## APIs
+
+### Request right-sizing recommendation v2
+
+The primary user of ContainerStats pipeline data is v2 of the request right-
+sizing recommendation API. ContainerStats data is used for quantile-based
+recommendations. [Docs](https://github.com/kubecost/docs/blob/main/api-request-right-sizing-v2.md)
+
+### Debugging
+
+There is an _unstable_, _unofficial_ API for introspecting pipeline data
+available at `/model/containerstats/quantiles`.
+
+## Behavior
+
+The pipeline builds 24 hour "windows" of data. It only builds complete windows,
+e.g. if now is `2003-08-23T08:30:00Z` the pipeline will only build up to the
+window from `2003-08-22T00:00:00Z` to `2003-08-23T00:00:00Z`.
+
+The pipeline will return an error response if a requested time range of data
+contains any windows (24 hour chunks) are "expected" (should be in the store)
+but not "available" (the pipeline has not yet built and loaded a complete set of
+data into the store).
+
+## Logs
+
+All ContainerStats-related log messages should contain `ContainerStats` or
+`ContainerStatsSet`. The pipeline logs a few things at INFO level to show that
+the pipeline is running successfully. Much greater detail is available at the
+DEBUG level. See [the official instructions](https://github.com/kubecost/cost-analyzer-helm-chart#adjusting-log-output) if you would like to learn how to change the log level.
+
+## Configuration
+
+The ContainerStats pipeline's behavior is controller by a few different
+environment variables.
+
+| Environment variable | Helm chart value | Description |
+|----------------------|------------------|-------------|
+| `CONTAINER_STATS_ENABLED` | `kubecostModel.containerStatsEnabled` | Enables the pipeline. |
+| Prometheus/Thanos configuration | | The pipeline inherits most of the existing Prometheus/Thanos configuration because it leverages the same client(s) used by the Asset and Allocation pipelines. Specific deviations will be mentioned. |
+| `ETL_MAX_PROMETHEUS_QUERY_DURATION_MINUTES` | ??? | The pipeline will obey this, but may fail to initialize if this is set below the minimum value supported by the pipeline (10 minutes).
+| "Storage" configration | | The pipeline inherits most of the existing "store" configuration used by other pipelines like Asset and Allocation. This includes, but is not limited to: store duration, store type (file, federated, etc.), leader election, storage pathing, storage directory, bucket storage, and backup. |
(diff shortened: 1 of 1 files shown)


michaelmdresser:
@nikovacevic do you think this info is okay to publicize? I think having at least _something_ would be helpful.

containerstats-pipeline.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Request right-sizing recommendation v2
+
+The primary user of ContainerStats pipeline data is v2 of the request right-
+sizing recommendation API. ContainerStats data is used for quantile-based
+recommendations. Review the doc for this feature [here](https://github.com/kubecost/docs/blob/main/api-request-right-sizing-v2.md).
+
+### Debugging
+
+There is an _unstable_, _unofficial_ API for introspecting pipeline data
```
bstuder99, line 16: I don't know how I feel about this emphasis on the word unstable. Would this API still be considered in beta, or in testing? I think there are better word choices.

rossfisherkc: I think it depends... if it's going to be something we push out as "beta", not fit for production usage and will sit unfinished for a long time, I think the emphasis is important, also, if there's any risk of us just dropping it randomly. If it's something we are committed to finishing and making stable, officially supporting within a sane timeframe, I agree that softer language is okay.

Can using the unfinished API cause any sort of downtime or data loss?

bstuder99: I can get behind informing the user that a feature is still being tested/developed, but calling it unstable may make them wary of even looking into it, so I think we should elaborate what the specific harm/faults could be here.

rossfisherkc: My backstory is that at my last job, we launched MySQL servers under "beta" and they sat with no further dev for 2-3 years, customers sometimes permanently lost data and many used it in production. Bad business decisions on behalf of said employer, for sure, but I did fight hard to include additional unstable/unsafe language.

Not that our product is as mission critical and can open the customer to critical data loss ;) Just wanted to share my thought process, there's a balance there somewhere and if this API could be suddenly dropped or deemed unfeasible in the future, totally cool with erring on the side of non production language, imo

michaelmdresser: My intent was to communicate that the API endpoint being discussed in this subsection (`/model/containerstats/quantiles`) does not have a **stable schema** and is **not supported as an official product feature**. Rather, it is a debugging-only endpoint mostly for use by support/solutions eng with users OR by eng during development.

Ideally, this would belong in an "internal-only" section of this document, but I'm not sure if we are able to do that right now.

For either/both of you, is there a rephrase you would suggest based on this additional context?

bstuder99: I think the way you put it, not being supported currently as an official product feature" is probably better than "unstable." Is there a reason this section needs to be included at all however, as it's mainly for internal use?

michaelmdresser: I'm not aware of an internal-only documentation location for this information that people actually check. Is there one?

containerstats-pipeline.md
```diff
@@ -19,20 +19,20 @@ available at `/model/containerstats/quantiles`.
...
 The pipeline builds 24 hour "windows" of data. It only builds complete windows,
-e.g. if now is `2003-08-23T08:30:00Z` the pipeline will only build up to the
+e.g. if now is `2003-08-23T08:30:00Z`, the pipeline will only build up to the
 window from `2003-08-22T00:00:00Z` to `2003-08-23T00:00:00Z`.
 
 The pipeline will return an error response if a requested time range of data
-contains any windows (24 hour chunks) are "expected" (should be in the store)
-but not "available" (the pipeline has not yet built and loaded a complete set of
+contains any windows (24 hour chunks) are expected (should be in the store)
+but not available (the pipeline has not yet built and loaded a complete set of
```
michaelmdresser, line 29: @bstuder99 the "expected" and "available" were quoted to highlight them because they match nomenclature in the code, log messages, and error messages for this feature set. What's the best way to represent that?

containerstats-pipeline.md
michaelmdresser: Also, updated phrasing.

containerstats-pipeline.md
bstuder99: Gotcha, I just assumed the quotations were meant to introduce new terminology, but figured the following parentheses clarify these terms perfectly and so the quotes were redundant. If these terms appear in log/error messages, we should just format them in italics.

containerstats-pipeline.md
bstuder99: I don't know about an internal doc specifically, I'm just saying if we are having difficult wording the description of a non-official feature, we should consider not documenting the feature at all. Otherwise, if we think it should still be added to the public doc, I word prefer your description in this thread as a clarifier as opposed to saying "unstable." However, this is close to a non-issue so I really don't think it needs more discussion.

bstuder99 (approved):

containerstats-pipeline.md
michaelmdresser: Thanks; italicized.

containerstats-pipeline.md
michaelmdresser: I updated the description to use what I said earlier in the thread instead of unstable.

srpomeroy:
@nikovacevic Are you ok with this information being public?

@rossfisherkc / @bstuder99 I am going to start a thread about internal docs via GitBooks.

nikovacevic:
I don't have an opinion on it being public or private, personally. In this case, I don't see any reason for users to need to use this API (correct me if I'm wrong, @michaelmdresser) so I would lean towards putting this same information in an internal doc.

That said, if we want to publish it, I don't see any harm. The word "experimental" seems more pertinent to me than "unstable" FWIW. Let me know if I can help beyond that.

nikovacevic:
e.g. Summary Allocation API is still not publicly documented AFAIK (https://docs.google.com/document/d/1W-HY5B-Hd1K-jelbAjI6qOvgRIdb5PEXQ5weWh22bX0/edit) and that drives (or used to drive) the Allocation page.

michaelmdresser:
Please let me know if we find a good way to colocate public and internal documentation -- I would have split this between something public and something private, but I guessed that the cost of doc fragmentation/lack of discoverability might be worse than what little we lose by publicizing.

bstuder99:
Well, Gitbooks will be a great place for getting an internal docs repo set up. Support has a library of internal content we'll be hosting, and we can always expand.

bstuder99:
Maybe create a plan for any internal doc if we should plan to eventually take it public

PR #437, opened 2022-11-09
Add authURL clarification in OIDC public docs

biancaburtoiu:
This PR adds a clarifying note on the `.Values.oidc.authURL` parameter that was previously missed from the internal docs and customers have asked about. 

It also downgrades the "Keycloak setup" header to fit under "Supported identity providers" as opposed to being at the same level.


srpomeroy (approved):
Lgtm

PR #444, opened 2022-11-14
Create http-troubleshooting.md

bstuder99:
Creation of a troubleshooting doc for HTTP codes based on @thomasvn's support doc. We can expand this over time using real customer questions/issues.

Do we need to provide contextual info of the hundreds digit with categories? Maybe as we expand (ie. all 400 errors are client, all 500 are servers, etc.) Only would bother adding this if most users aren't super familiar with HTTP status codes as a concept.


bstuder99:
I've also surmised the most recent issue which prompted the creation of this doc in the Gateway timeout section, can I get some eyes on that part in particular?

http-troubleshooting.md
```diff
@@ -0,0 +1,32 @@
+# Kubecost HTTP Response Code Troubleshooting
+
+If you have experienced an HTTP response code while using Kubecost, consult this doc for a possible fix.
+
+You will likely encounter these response codes in the browser, when performing REST requests through a client like cURL, or when viewing the logs `kubectl logs deploy/kubecost-cost-analyzer -c cost-analyzer-frontend -n kubecost`.
```
bstuder99, line 5: In Thomas' doc, this is all written as one singular log, is it supposed to several? If it's just the one, I need to correct the plural "logs".

bstuder99:
@thomasvn @jessegoodier Can I get a review on this?

jessegoodier (dismissed):
lgtm

http-troubleshooting.md
```diff
@@ -0,0 +1,37 @@
...
+
+## HTTP 403 Forbidden
+
+This is most likely due to a user authenticating to Kubecost or performing an action in Kubecost that they don't have permissions to do (for example, saving a report when the user has read-only access). It can be caused by mesh networking, firewalls, load balancers, or some other service
```
thomasvn, line 9: There is lower likelihood of this error being caused by "mesh networking, firewalls, load balancers, or some other service"

It could be safer to make the following statement:
> "There is also a possibility that this error is caused by intermediary services such as load balancers, firewalls, or service meshes"

http-troubleshooting.md
```diff
@@ -0,0 +1,37 @@
...
+### Test command
+
+> **Note**: The following test command can be used for troubleshooting both 499 and 504 errors.
+
+* If running the following command fails or hangs when the pod is ready, the error is likely due to intermittent DNS:
+  *  `kubectl exec -i -t -n kubecost kubecost-cost-analyzer-55c45d9d95-8m2sq -c cost-analyzer-frontend -- curl kubecost-cost-analyzer.kubecost:9090/model/clusterInfo`
+
+## HTTP 504 Gateway timeout
+
+Almost always because the `cost-model` container in the `kubecost-cost-analyzer` pod is down. If the pods are consistently restarting, this could be due to OOM (Out of Memory) errors. As noted in our [AWS Cloud Integration](https://guide.kubecost.com/hc/en-us/articles/4407595928087-AWS-Cloud-Integration#summary-and-pricing) doc:
```
thomasvn, line 32: Update link to doc: `[AWS Cloud Integration](./aws-cloud-integrations.md)`

thomasvn:
Made a few comments. Besides that I think this is a good initial version to release!

thomasvn (approved):

PR #445, opened 2022-11-15
add new instruction to install Kubecost operator

linhlam-kc:
Signed-off-by: Linh Lam <linh@kubecost.com>

Add new instruction to install Kubecost operator from Red Hat OpenShift web console.


bstuder99 (approved):
LGTM

PR #446, opened 2022-11-17
Add documentation for continuous request right-sizing

michaelmdresser:
- Pairs with https://github.com/kubecost/cost-analyzer-helm-chart/pull/1804


michaelmdresser:
@dwbrown2 do you think a blog post on top of this is warranted?

bstuder99:
The exclamation marks feel a bit excessive and too enthusiastic for a tech doc, but a blog post would def be more suitable. A misc. note, how do we usually format "cluster controller"? I believe as a product feature, we currently have it capitalized in our docs as opposed to just writing cluster-controller, but is there a context in which one is right over the other? Probably would only go with the lowercase/hyphenated form when referring to code, or the specific Git location.

michaelmdresser:
Thanks for the updates. I just re-capitalized "Deployments" because it is used in a proper noun K8s context here. I also moved the filename to better reflect the title update.

> how do we usually format "cluster controller"? 

I don't think there is consensus here. Certainly agree that in a Git/GitHub context "cluster-controller" makes sense because it is the name of the repo. In the official doc in this repo (`controller.md`) is is indeed currently capitalized as "Cluster Controller". It can possibly get ambiguous in the context of deploying and running Kubecost because the Deployment is called `cluster-controller`, just like the repo and container.

For the sake of consistency in docs, I've updated this new doc to use "Cluster Controller". If you think there's a better approach @bstuder99, I'd be happy to follow it!

bstuder99:
It doesn't come up in docs too much, but I think Kubecost product features should try to be capitalized where relevant. We had a discussion a month or two back about certain Kubernetes concepts and when to capitalize them, and Kubernetes as an org is somewhat inconsistent about it (words like pod are capitalized or lowercase depending on the article). With the word deployment, it feels as though this is another example, I've seen it capitalized but for seemingly no other reason than it is the subject of the article it appears in. Not a big deal for now as long as we're consistent and add it to our tech vocab section of the style guide.

bstuder99 (approved):

PR #447, opened 2022-11-18
cpu limits for network-costs

jessegoodier:
Please review.

### DIFF
--- a/network-allocation.md
+++ b/network-allocation.md
@@ -23,7 +23,7 @@ To enable this feature, set the following parameter in _values.yaml_ during [Hel
 
 > **Note**: Network cost, which is disabled by default, needs to be run as a privileged pod to access the relevant networking kernel module on the host machine.
 
-## Resource Limiting
+## Resource limiting
 
 In order to reduce resource usage, Kubecost recommends setting a CPU limit on the network-costs daemonset. This will cause a few seconds delay during peak usage and does not effect overall accuracy. This is done by default in Kubecost 1.99+.
 
--- a/network-allocation.md
+++ b/network-allocation.md
@@ -10,12 +10,14 @@ Metrics include egress and ingress data transfers by pod and are classified as i
 ## Enabling network costs
 
 To enable this feature, set the following parameter in _values.yaml_ during [Helm installation](http://kubecost.com/install):
+
  ```
  networkCosts.enabled=true
  ```
+
  You can view a list of common config options [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/700cfa306c8e78bc9a1039b584769b9a0e0757d0/cost-analyzer/values.yaml#L573).
 
- If using the included Prometheus instance, the scrape is automatically conifgured.
+ If using the included Prometheus instance, the scrape is automatically configured.
 
  If you are integrating with an existing Prometheus, you can set `networkCosts.prometheusScrape=true` and the network costs service should be auto-discovered.
 
@@ -29,7 +31,7 @@ In order to reduce resource usage, Kubecost recommends setting a CPU limit on th
 
 For existing deployments, these are the recommended values:
 
-```
+```yaml
     resources:
       limits:
         cpu: 500m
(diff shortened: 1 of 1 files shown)


AjayTripathy (approved):

dwbrown2:
There's a lot of discussion about cpu limits -- https://home.robusta.dev/blog/stop-using-cpu-limits

Do we want to actively recommend?

network-allocation.md
```diff
@@ -5,16 +5,39 @@ This document summarizes Kubecost network cost allocation, how to enable it, and
...
+
+ Alternatively a serviceMonitor is also [available](https://github.com/kubecost/cost-analyzer-helm-chart/blob/700cfa306c8e78bc9a1039b584769b9a0e0757d0/cost-analyzer/values.yaml#L716).
+
 > **Note**: Network cost, which is disabled by default, needs to be run as a privileged pod to access the relevant networking kernel module on the host machine.
- 
+
+## Resource Limiting
+
+In order to reduce resource usage, Kubecost recommends setting a CPU limit on the network-costs daemonset. This will cause a few seconds delay during peak usage and does not effect overall accuracy. This is done by default in Kubecost 1.99+.
+
```
bstuder99, line 29: Are we holding off on merging this until 1.99 then?

bstuder99 (approved):

network-allocation.md
bstuder99: If so, we should hold off on merging and tag this PR

mbolt35:
> There's a lot of discussion about cpu limits -- https://home.robusta.dev/blog/stop-using-cpu-limits
> 
> Do we want to actively recommend?

The author does a great job explaining (both via the analogy AND the technical writeup). Although, he's looking at this from the "owner" of a cluster. In the context of running your own infrastructure and understanding how kubernetes works when requests and limits are set (or not set), then I 100% agree with this post. However, when our feedback from a customer is that "network costs is using a lot of CPU", we only have a few options. I would love to provide them information about how to restrict the network-costs CPU with requests, which will maximize their CPU usage (especially if there is excess available). I think "high CPU" gets misunderstood, like the article mentions, as "it's taking CPU away from other pods/containers." If we try to provide a corrective measure or point our users towards more information, and it isn't well received, then CPU limits does feel like the only remaining tool to guarantee that a pod doesn't exceed some threshold. 

I do feel strongly we should be guiding our users towards proper kubernetes management, but when this fails, it does seem reasonable to leverage less favorable options to ensure that their needs are met. 

@jessegoodier You may want to just provide a note referencing the content (or article) @dwbrown2 mentioned. I can confirm if the pod is hard limited, then it should function appropriately as long as it is not completely starved, so setting limits will work if the only goal is to prevent the pod from using "high CPU."

jessegoodier:
> There's a lot of discussion about cpu limits -- https://home.robusta.dev/blog/stop-using-cpu-limits
> 
> Do we want to actively recommend?

I have read that- very interesting.

I think this case may be the exception to the article.

My thinking is that a workload that is not latency sensitive (like a cumulative metric endpoint) can be CPU limited with no impact to the value the workload provides. 

The alternative is that no-limits will cause actual workloads that are latency sensitive to compete for CPU resources exactly when the system is under load.

And from a support perspective we have customers that are noticing that this is an expensive workload- questioning the value of running it.

I'm not an expert here, so totally open to killing this change.

network-allocation.md
jessegoodier: I'll have a draft PR for the change in the helm chart. No harm in publishing this doc now given we have instructions for implementing the change.

network-allocation.md
```diff
@@ -5,16 +5,40 @@ This document summarizes Kubecost network cost allocation, how to enable it, and
...
- 
+ You can view a list of common config options [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/700cfa306c8e78bc9a1039b584769b9a0e0757d0/cost-analyzer/values.yaml#L573).
+
+ If using the included Prometheus instance, the scrape is automatically conifgured.
+
+ If you are integrating with an existing Prometheus, you can set `networkCosts.prometheusScrape=true` and the network costs service should be auto-discovered.
+
+ Alternatively a serviceMonitor is also [available](https://github.com/kubecost/cost-analyzer-helm-chart/blob/700cfa306c8e78bc9a1039b584769b9a0e0757d0/cost-analyzer/values.yaml#L716).
+
 > **Note**: Network cost, which is disabled by default, needs to be run as a privileged pod to access the relevant networking kernel module on the host machine.
```
jessegoodier, line 24: @dwbrown2 any changes that you'd recommend?

network-allocation.md
```diff
@@ -5,16 +5,40 @@ This document summarizes Kubecost network cost allocation, how to enable it, and
...
+
+```
+    resources:
+      limits:
+        cpu: 500m
+      requests:
+        cpu: 50m
+        memory: 20Mi
+```
+ > This write-up regarding CPU limiting <https://home.robusta.dev/blog/stop-using-cpu-limits> has points worth considering. In the case of low-priority/latency insensitive jobs, there is little risk to setting CPU limits.
```
jessegoodier, line 40: any changes that you'd recommend?

jessegoodier: I removed the note.

I think it causes more confusion than clarity. 

I'm open to adding more color if we think it is needed.

PR #453, opened 2022-11-21
Remove legacy docs platform code

srpomeroy:
THese files are a holdover from the original GitHub Pages docs site. They should no longer be required.


srpomeroy:
@dwbrown2 Any reason not to remove these items?

dwbrown2:
@srpomeroy you guys have GA connected already yeah? If so, think you can get rid of the majority. 

I believe the `cname` file was a requirement of GH pages. Not 100% but that's the only one I'm not sure of.

srpomeroy:
@dwbrown2 `CNAME` is definitely for GitHub Pages.
https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain

dwbrown2:
Thanks for confirming, think all are good to go then assuming GA is set.

PR #456, opened 2022-11-22
Create kubecost-cloud-architecture.md

bstuder99:
Creation of the first of hopefully many Kubecost Cloud docs. For this doc, I think we should add an additional paragraph in the Architecture Overview section explaining what interactions are displayed in the graphic.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -102,6 +102,7 @@
 ## Architecture
 
 * [Kubecost Core Architecture Overview](architecture.md)
+* [Kubecost Cloud Architecture Overview](kubecost-cloud-architecture.md)
 * [Open Source](open-source-deps.md)
 * [Security and Data Protection](security.md)
 * [Ports](ports.md)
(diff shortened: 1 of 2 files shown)


srpomeroy:
@bstuder99 I pushed a commit for updating the nav.

jessegoodier:
What do you think of these changes?
also, @srpomeroy 
image

bstuder99:
Also wondering if we needed to actually label Prometheus? Will the logo be universally understood?

jessegoodier:
> Also wondering if we needed to actually label Prometheus? Will the logo be universally understood?

Maybe make it much smaller? I good with or without it.

bstuder99:
Ok, the chart has been replaced in kubecost/docs/images, so no changes on this doc have to be made. The link has already updated.

jessegoodier (approved):
can you add an s to kube-state-metrics? looks good.

PR #457, opened 2022-11-23
add note for hourly windows

keithhand:
I tested myself and this still seems to hold true. Any hourly windows that can't be satisfied by a daily window return:

> Recommending requests from data: CPU algorithm set to max but no max data is available

Doesn't Work:
- https://demo.kubecost.io/model/savings/requestSizingV2?window=1h
- https://demo.kubecost.io/model/savings/requestSizingV2?window=3h
- https://demo.kubecost.io/model/savings/requestSizingV2?window=12h

Works (same data):
- http://demo.kubecost.io/model/savings/requestSizingV2?window=1d
- http://demo.kubecost.io/model/savings/requestSizingV2?window=24h
- http://demo.kubecost.io/model/savings/requestSizingV2?window=36h

Closes https://github.com/kubecost/support/issues/26.


bstuder99 (approved):

srpomeroy (approved):
LGTM

PR #464, opened 2022-11-30
Match helm chart description

jcharcalla:
From support issue: https://github.com/kubecost/support/issues/29

Updated description to match the helm chart. https://github.com/kubecost/cost-analyzer-helm-chart/blob/a998887abf81bdf35c34e429293bb1b12bcbe3dc/cost-analyzer/values.yaml#L852

### DIFF
--- a/aws-cloud-integrations.md
+++ b/aws-cloud-integrations.md
@@ -416,7 +416,7 @@ kubecostProductConfigs:
 
 These values can either be set from the kubecost frontend or via .Values.kubecostProductConfigs in the helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be overridden on pod restart.
 
-* `athenaProjectID` e.g. "530337586277" # The AWS AccountID where the Athena CUR is.
+* `athenaProjectID` e.g. "530337586277" # The AWS AccountID where the Athena CUR is. Generally your masterpayer account.
 * `athenaBucketName` An S3 bucket to store Athena query results that you’ve created that Kubecost has permission to access
   * The name of the bucket should match `s3://aws-athena-query-results-*`, so the IAM roles defined above will automatically allow access to it
   * The bucket can have a Canned ACL of `Private` or other permissions as you see fit.


bstuder99:
Can't find anything on the term masterpayer and how it should be formatted, but I did get some misc. AWS docs to come up. Is this similar to "master/management account"?

jcharcalla:
@bstuder99 Good point, it's both "master payer" and "masterpayer" on that same page https://github.com/kubecost/docs/blob/jcharcalla-multicloud-clarification/aws-cloud-integrations.md

However, This [AWS document](https://aws.amazon.com/organizations/faqs/) states "the management account (formerly known as master account)" implies it may be the "management account".

From google this looks like the previous definition.

> What is AWS master payer account?
This account has the role of a payer account and is responsible for paying all charges accrued by the accounts in its organization. You cannot change which account in your organization is the management account.

As opposed to whats now in the FAQ for 

> Q: What is a management account (formerly known as master account)?

Sounds like we will need to verify the correct terminology and update docs accordingly.

bstuder99:
I'm fine with "master payer", long as we're consistent it should be okay. If we want to update to management payer, it should take about five mins.

aws-cloud-integrations.md
```diff
@@ -416,7 +416,7 @@ kubecostProductConfigs:
 
 These values can either be set from the kubecost frontend or via .Values.kubecostProductConfigs in the helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be overridden on pod restart.
 
-* `athenaProjectID` e.g. "530337586277" # The AWS AccountID where the Athena CUR is.
+* `athenaProjectID` e.g. "530337586277" # The AWS AccountID where the Athena CUR is. Generally your masterpayer account.
```
bstuder99, line 419:
```suggestion
* `athenaProjectID` e.g. "530337586277" # The AWS AccountID where the Athena CUR is. Generally your master payer account.
```

bstuder99 (approved):

PR #471, opened 2022-12-05
Move auto (1-click and continuous) request sizing setup to dedicated doc

michaelmdresser:
Provides a single point to link to without eliminating more specific documentation for 1-click and continuous request right-sizing.


michaelmdresser:
Supports the single "Set Up" button in the FE. See https://github.com/kubecost/cost-analyzer-frontend/pull/1397

bstuder99:
So this is consolidating content from multiple docs into one, but no docs are being deleted here, right?

bstuder99:
Just curious if we have to change anything manually in summary.md if it affects navigation in any way

michaelmdresser:
Aha, I didn't know about summary.md, thanks. I added a commit that updates summary.md, hopefully that's good

michaelmdresser:
> So this is consolidating content from multiple docs into one, but no docs are being deleted here, right?

Basically correct. Generic setup info is being centralized into a new generic doc, while leaving specific guide/usage docs alone.

bstuder99:
Took me a minute to wrap my head around this after viewing its structure in summary.md. So the Automatic Request doc is completely new and the Continuous Request is being moved into the docs nav?

bstuder99 (approved):

michaelmdresser:
> So the Automatic Request doc is completely new and the Continuous Request is being moved into the docs nav?

"Auto request sizing" is a parent description/setup doc for two features: Continuous and 1-click. Each of those features has its own doc and Continuous is new as of v1.99. The two features have the same setup instructions and overall theme, so we needed a single setup doc to link to from the frontend.

If it still isn't clear, that's probably a sign that the docs don't read very well! I'm happy to work on improvements; what's causing you confusion?

> Are these all 1.99 features/updates?

Yes.

michaelmdresser:
Going to merge, happy to keep working on improving clarity in a future PR!

bstuder99:
I didn't think the content itself was confusing, I was more curious about how docs are being created/deleted/moved around. I will take another cursory look though for where clarity could be added.

PR #474, opened 2022-12-06
Update FAQ describing multiaggregation on the Allocation API.

thomasvn:
This is the recommended workaround when users are trying to solve the following problem:

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1839


srpomeroy (approved):

srpomeroy:
This is super helpful. I've stumbled across this a few times!

PR #476, opened 2022-12-08
Updated to discuss OpenCost

mattray:
Also updated go.mod link, probably need to include front-end deps too once the UI is being deployed

### DIFF
--- a/open-source-deps.md
+++ b/open-source-deps.md
@@ -1,9 +1,7 @@
 Open-Source
 ===========
 
-Kubecost has open source at the center of our business model. 
+Kubecost has open source at the center of our business model (open-core).
 
-Our commercial project is built on top of the [OpenCost CNCF Sandbox project](https://www.cncf.io/projects/opencost/), which has an Apache 2.0 license. 
+Our commercial product is built on top of the [OpenCost CNCF Sandbox project](https://www.cncf.io/projects/opencost/), which has an Apache 2.0 license.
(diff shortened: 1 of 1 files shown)


open-source-deps.md
```diff
@@ -1,9 +1,9 @@
 Open-Source
 ===========
 
-Kubecost has an open core business model. 
+Kubecost has open source at the center of our business model.
```
jessegoodier, line 4: open-core is a specific business model. I would at least call that out?

like:
```
Kubecost has open source at the center of our business model (open-core). 
```

But I agree that it was previously vague and assumed the reader understood open-core.

open-source-deps.md
```diff
@@ -1,9 +1,9 @@
 Open-Source
 ===========
 
-Kubecost has an open core business model. 
+Kubecost has open source at the center of our business model. 
 
-Our commercial project is based on our cost-model open-source project, which has an Apache 2.0 license. 
-Here is a list of the direct [open-source dependencies](https://github.com/kubecost/cost-model/blob/master/go.mod#L9) that we also include as part of our project.
+Our commercial project is built on top of the [OpenCost CNCF Sandbox project](https://www.cncf.io/projects/opencost/), which has an Apache 2.0 license.
```
dwbrown2, line 6: Commercial products or product instead of project?

open-source-deps.md
srpomeroy: Agreed, added back open-core.

open-source-deps.md
srpomeroy: Yep!

PR #477, opened 2022-12-08
CUR config clarity

jcharcalla:
Clarifying what options are required when creating the CUR report, additionally added the error one would expect to see when resource IDs are not included in the report.


bstuder99 (approved):

jcharcalla:
@srpomeroy No I did not, thanks for catching that.

PR #486, opened 2022-12-15
Do not Merge - Draft PR as first pass of Getting Started

teevans:
Signed-off-by: Thomas Evans <tevans3@icloud.com>


thomasvn:
Love it! I think this is a good step in refining the day1 experience for the user when opening Kubecost docs

dwbrown2:
Wow, love it!! Happy to weigh in specific copy if useful. Kai and I are planning to do a deep dive here in Jan.

bstuder99:
What's the status here? Some good stuff in here like Kubecost v OpenCost, but I think this is now pretty heavily outdated with things like the nav. I can port over this if we want to move forward, otherwise I'll close it. Low priority, just want to keep the repo clean.

PR #488, opened 2022-12-15
Remove ZD Metadata

srpomeroy:
This PR removes the ZD metadata that is no longer required.
It also removes the disabled GitHub action to sync docs.


thomasvn (approved):
LGTM!

PR #491, opened 2022-12-21
Update allocation.md to account for querying the cost-model Pod directly

nickolashkraus:
Had to dig around for why `/model/allocation/compute` does not work when querying the cost-model container.

Only the  `/allocation` handler is defined:

```go
a.Router.GET("/allocation/compute", a.ComputeAllocationHandler)
```

An NGINX location block is defined for `/model/` with a [proxy_pass](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass) defined:

```
location /model/ {
    ...
    proxy_pass http://model/;
}
```

>If the proxy_pass directive is specified with a URI, then when a request is passed to the server, the part of a [normalized](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) request URI matching the location is replaced by a URI specified in the directive...

### DIFF
--- a/allocation.md
+++ b/allocation.md
@@ -6,7 +6,7 @@ The Allocation API is the preferred way to query for costs and resources allocat
 http://<kubecost>/model/allocation
 ```
 
-> **Note**: Throughout, we use `localhost:9090` as the default Kubecost URL, but your Kubecost instance may be exposed by a service or ingress. To reach Kubecost at port 9090, run: `kubectl port-forward deployment/kubecost-cost-analyzer -n kubecost 9090`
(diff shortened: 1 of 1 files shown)


thomasvn:
Hey @nickolashkraus, thanks for the PR!

Just to clarify, are you saying that when doing the following, you will need to omit the "/model" from the URI?

```bash
# Get shell in cost-analyzer-frontend container
kubectl exec -it deployment/kubecost-cost-analyzer -c cost-analyzer-frontend sh

# Make HTTP requests to the cost-model backend
curl localhost:9003/allocation/compute?window=1d
```

If so, I agree with your finding and also agree that this is definitely something worth documenting. Thanks!

thomasvn (approved):

nickolashkraus:
@thomasvn That is correct. This note would apply for those only using the Kubecost API via the cost-model container or are deploying OpenCost using the [OpenCost installation](https://www.opencost.io/docs/install).

PR #496, opened 2023-01-03
kubecost cloud docs

jessegoodier:
Add missing line break in list
Consistent wording for limited access vs alpha. 

Please verify TOC updates are consistent?


bstuder99:
I think the Alpha is a carryover from when this article was called Kubecost Hosted. Have we moved out of limited availability for Kubecost Cloud?

PR #507, opened 2023-01-06
Include Helm value in federated-etl.md

thomasvn:
Update federated-etl.md to include corresponding Helm value which must be set


thomasvn:
Made the following finding with @keithhand @mmurph3. Adding for review.

srpomeroy (dismissed):
LGTM

thomasvn:
@jessegoodier Good catch! I agree that each federated cluster requires a unique `cluster_id`. I've pushed a new commit.

Requesting another review @srpomeroy !

kaelanspatel (approved):
lgtm fwiw

Very slight caveat (no changes needed) but only `cluster_id` is needed for federated to work.

PR #511, opened 2023-01-13
Renaming Filters V2 doc

bstuder99:
Renames the v2 Filters article title and file name to be more descriptive, corrects summary.md for both changes.


thomasvn:
@bstuder99 This looks good! Renaming the title here adds more clarity.

One thing to keep an eye on in the future though, is that these V2 Filters will eventually be used in places beyond the Allocation API. For example, V2 Filters are currently available for use on the "/request-sizing" view in the Beta UI.

bstuder99:
Good catch. The language in the doc was a bit unclear and alluded to this filtering language becoming more accessible over time. Let me go back and make edits with this in mind. Any other APIs we plan on adding this functionality to?

srpomeroy:
Should we consider listing what APIs v2 filters work with? Or perhaps link to it from the relevant pages/apis?

bstuder99:
I've already added a section within the v2 doc which links the compatible APIs. I'm still working on the Allocation API page but I will add a section explaining the advanced filtering language.

filters-api.md
```diff
@@ -1,10 +1,10 @@
-# V2 Filters
+# API Filters (v2)
```
thomasvn, line 1: Since this isn't strictly an API, maybe we can title it something like "Filter Parameters (v2)"? Thoughts?

SUMMARY.md
```diff
@@ -93,7 +93,7 @@
 
 * [Kubecost API Directory](apis-overview.md)
   * [Allocation API](allocation.md)
-  * [V2 Filters](filters-v2.md)
+  * [Allocation API Filters (v2)](filters-api.md)
```
thomasvn, line 96: Reminder to change `summary.md` once you change the other doc's title!

Also since V2 filters doesn't apply solely to the Allocation API anymore ... should this doc be moved to the bottom of the Kubecost API Directory? (instead of being directly under the Allocation API doc)

srpomeroy: Adding linting doc titles vs summary.md to the backlog for PR checks. 😁

bstuder99:
Ok, to clarify:
* Article title is now Filter Parameters (v2)
* MD file is now filters-api.md
* File name and title have been corrected in Summary.md, being moved to the bottom of the API nav
* Links to both APIs which support the filtering language have been added to the article

thomasvn:
@bstuder99 `summary.md` looks good, but `filters-api.md` doesn't have the updated article title?

bstuder99:
So I don't think we have to manually update article titles inside the Markdown file itseld, just how the article title is displayed in Summary.md. From my trial and error, I think Summary.md has the final say on article titles (tried to manually change an article in GitBook to no avail).

So in this case,  * [Filter Parameters (v2)](filters-api.md) in Summary.md should be all we need, assuming the .md file names align (which they do).

srpomeroy:
Correct. If the summary and the doc title don't match, then the summary wins. However, we should try to keep them in sync if possible.

thomasvn:
Ah! Learned something new. Thanks @bstuder99 & @srpomeroy !

thomasvn (approved):

PR #514, opened 2023-01-18
productKey config not required on secondary clusters

thomasvn:
* Make explicit that you don't need to add productKey config on secondary clusters
* Also did some markdown linting

### DIFF
--- a/secondary-clusters.md
+++ b/secondary-clusters.md
@@ -1,6 +1,6 @@
 # Secondary Clusters Guide
 
-Secondary clusters use a minimal Kubecost deployment to send their metrics to a central storage-bucket (aka durable storage) that is accessed by the primary cluster to provide a ***single-pane-of-glass*** view into all aggregated cluster costs globally. This aggregated cluster view is exclusive to Kubecost Enterprise.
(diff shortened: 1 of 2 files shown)


secondary-clusters.md
```diff
@@ -1,13 +1,12 @@
-Secondary Clusters Guide
-===================
+# Secondary Clusters Guide
 
-Secondary clusters use a minimal Kubecost deployment to send their metrics to a central storage-bucket (aka durable storage) that is accessed by the primary cluster to provide a ___single-pane-of-glass___ view into all aggregated cluster costs globally. This aggregated cluster view is exclusive to Kubecost Enterprise.
+Secondary clusters use a minimal Kubecost deployment to send their metrics to a central storage-bucket (aka durable storage) that is accessed by the primary cluster to provide a ***single-pane-of-glass*** view into all aggregated cluster costs globally. This aggregated cluster view is exclusive to Kubecost Enterprise.
```
bstuder99, line 3: Is it necessary to do both bold and italics for this term? I would go with bold as it is some unique vocab but not an interactable element.

bstuder99:
Thank you for fixing up some of those links and using the new tagging system. I have a minor tweak but aside from that, this looks good.

bstuder99 (dismissed):

PR #521, opened 2023-01-23
Note why uncontrolled pods cannot be supported in Kubescaler/Continuous request right sizing

michaelmdresser:
There will be another update to this doc before the release of 1.100 that increases the size of the "supports" section, just want to open this PR so I don't forget the link.


michaelmdresser:
Merging so I can branch

PR #536, opened 2023-01-30
Update openshift-operatorhub-kubecost-install.md

linhlam-kc:
Update example CRD templates to deploy Kubecost v1.99 that supports OCP 12.0 (K8s 1.25)

### DIFF
--- a/openshift-operatorhub-kubecost-install.md
+++ b/openshift-operatorhub-kubecost-install.md
@@ -63,77 +63,202 @@ kubectl apply -f example-crd.yaml -n kubecost
 apiVersion: charts.kubecost.com/v1alpha1
 kind: CostAnalyzer
 metadata:
-  name: kubecost
+  name: cost-analyzer-ocp-sample
 spec:
-  # Default values copied from <project_dir>/helm-charts/cost-analyzer/values.yaml
-  openshiftDeployment: true
   affinity: {}
-  awsstore:
-    createServiceAccount: false
-    useAwsStore: false
-  clusterController:
-    enabled: false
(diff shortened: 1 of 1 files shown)


openshift-operatorhub-kubecost-install.md
```diff
@@ -63,77 +63,202 @@ kubectl apply -f example-crd.yaml -n kubecost
...
+  kubecostNetworkCosts:
+    securityContext: {}
+    containerSecurityContext: {}
+
   prometheus:
     nodeExporter:
       enabled: false
-    serviceAccounts:
-      nodeExporter:
-        create: false
```
srpomeroy, line 87: @linhlam-kc Won't removing the service account block cause the node exporter service account to be created?

linhlam-kc: @srpomeroy Disabling nodeExporter should be enough according to this [condition](https://github.com/kubecost/openshift-helm-chart/blob/bc0e157f8d9376f93aaebe60ae300aaa0305c21a/cost-analyzer/charts/prometheus/templates/node-exporter-serviceaccount.yaml#L2) I do not see it is created in my testing last week as well on the OCP cluster. Let me try it real quick again.

linhlam-kc: I can confirm that disable promethus.nodeExporter.enabled will also disable the creation of nodeExporter serviceAccount

srpomeroy: Awesome! There must have been a change. That was not the case previously.

srpomeroy (approved):
LGTM

PR #543, opened 2023-02-07
Kubecost v1.100 will require Helmv3.1+

thomasvn:
Specifically, because v1.100 uses the `lookup` function [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.100/cost-analyzer/templates/NOTES.txt#L4-L8), which only works in Helm v3.1+.

### DIFF
--- a/alibaba-install.md
+++ b/alibaba-install.md
@@ -2,7 +2,7 @@
 
 ## Helm install Kubecost
 
-Kubecost installation is exactly same as other cloud providers with Helm 3:
+Kubecost installation is exactly same as other cloud providers with Helm v3.1+:
 
 ` helm install kubecost/cost-analyzer -n kubecost -f values.yaml`
 
--- a/install.md
+++ b/install.md
@@ -77,13 +77,13 @@ For larger teams and companies with more complex infrastructure, you need the ri
 
 ## Alternative installation methods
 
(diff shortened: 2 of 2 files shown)


thomasvn:
@linhlam-kc I noticed that `aws-amp-integration.md` and `aws-eks-cost-monitoring.md` state that Helm v3.9+ is required? Is that true? Do you think earlier versions of Helm are supported as well? Not urgent.

bstuder99:
Ok, I have an active PR for our 3-step install guide as well which mentions Helm version. Still need a confirmation on that Helm 3.9+ stuff

bstuder99 (approved):

thomasvn:
@bstuder99 Thanks for updating on the Kubecost website too!

linhlam-kc:
> @linhlam-kc I noticed that `aws-amp-integration.md` and `aws-eks-cost-monitoring.md` state that Helm v3.9+ is required? Is that true? Do you think earlier versions of Helm are supported as well? Not urgent.

For AWS integration specifically Helm 3.9+ is required to support Helm chart hosted on OCI registry. The earlier version does not work in my testing.

thomasvn:
@linhlam-kc Got it, thanks for verifying!

dwbrown2:
@AjayTripathy was this dependency change intentional? What % of users on <= 3.0?

AjayTripathy:
It is intentional and was discovered in testing.

I'm keeping a close eye on support. 3.1 is ~2yo at this point and there have been quite a few security vulns since then so I'm guessing very few in prod. May be more random users using it, but as noted we'll catch in support.

PR #544, opened 2023-02-07
federation updates

jessegoodier:
Summary:
1. changing federation doc to overview the options
2. move thanos setup to own doc
3. add detail/diagram to federated-etl doc
4. update multi-cluster options to differentiate editions

@bstuder99 Do you mind checking the nav/formatting/image links? I realize that is asking a lot, feel free to push back. This is not super time sensitive.

@srpomeroy Do you mind doing a full review? There are a lot of changes.

All others, feedback welcome!

PR #551, opened 2023-02-10
Move all multi-cluster docs under "Installation" in navbar

thomasvn:
Amends https://github.com/kubecost/docs/pull/544


srpomeroy (approved):

bstuder99:
Is there a PR showing where we merged all CSP long storage docs together? They're currently all still separated on the FE

thomasvn:
@bstuder99 Hm, I don't think we ever merged the AWS, GCP, and Azure docs together. Well, at least I don't recall.

bstuder99:
Sorry, I was looking at the files changed and the way it was depicting those files in red made me assume they were all merged also

PR #553, opened 2023-02-11
More clarity on efficiency and idle. Include mathematical definitions.

thomasvn:


bstuder99:
I tried to correct this last month. It gets confusing when written in plain text, but you have an odd number of total parenthesis so I don't believe this change can be correct. Can you double check? I can also DM you the Slack thread where we discussed this

efficiency-idle.md
```diff
@@ -4,23 +4,32 @@ For teams interested in reducing their Kubernetes costs, it's beneficial to firs
...
+Cluster idle cost is defined as the difference between the cost of allocated resources and the cost of the hardware they run on. Allocation is defined as the max of usage and requests. It can also be expressed as follows:
 
-So, idle costs can also be thought of as the cost of the space that the Kubernetes scheduler could add pods without disrupting any workloads in but is not currently. Idle can be charged back to pods on a cost-weighted basis or viewed as a separate line item.
+> *idle_cost = sum(node_cost) - (cpu_allocation_cost + ram_allocation_cost + gpu_allocation_cost)*
+>
+> where \
+> *allocation = max(request, usage)*
 
-As an example, consider the following representations:
+So, idle costs can also be thought of as the cost of the space that the Kubernetes scheduler could add pods without disrupting any workloads in but is not currently.
```
srpomeroy, line 28: This is confusing to me. How about if we reword it to something like this:
```suggestion
So, idle costs can also be thought of as the cost of the space that the Kubernetes scheduler could schedule pods, without disrupting any existing workloads, but it is not currently.
```

srpomeroy (approved):
One minor suggestion, otherwise LGTM

efficiency-idle.md
thomasvn: Good suggestion, merged!

thomasvn:
@srpomeroy Requesting a review whenever you get a moment!

srpomeroy (approved):
LGTM!

PR #556, opened 2023-02-16
Relocates image files in docs repo into images folder

bstuder99:
Will also fix all paths related to these images to avoid broken image links


bstuder99:
The three affected pages are custom-prom.md, aws-service-account-thanos.md, and context-switcher.md. I've checked and made sure all links work, but may need someone to double check. Two example yaml files are also moved over into images.

thomasvn:
@bstuder99 Seems like you used the path `kubecost/docs/images/images/` but I think it should just be `kubecost/docs/images/`?

Also, we currently have images stored in the `.gitbook/assets` folder. Maybe in the future we merge all images to one location?

thomasvn (approved):

PR #559, opened 2023-02-22
Update Navbar "Using Kubecost" section

thomasvn:
* ensure the core concepts are presented first
* advanced concepts presented lower


bstuder99:
I like this a lot. We may have a few additions to the UI sidebar coming up, and I think we should arrange the corresponding docs in the same order. So in this way, do you think it makes sense to add Cluster Health Score above Alerts? Also need to make a note to add an example screenshot into that doc.

thomasvn:
@bstuder99 100% agree. I'll make the change now

bstuder99 (approved):

PR #566, opened 2023-02-24
Add docs for using Storj as S3 bucket for ETL Backup / Thanos long-term storage solution

kaelanspatel:
Adds docs for the configuration and usage of Storj for backing up ETL/using as Thanos store. Storj claims S3 compatibility, so this technically just means we support configuring a Storj bucket as an S3 bucket. No additional configs are necessary. 

Verified that Storj can be used with all existing S3-based functions currently, as well as actual functionality w/ ETL backup and Federated ETL.


bstuder99:
It looks like this content is just repeated across both docs. I would say we should find one place for it to stay. Is Storj significant enough on the topic of long term storage to justify its own article (not saying it isn't), but if so, I think the way to implement this would be to hyperlink to it from the ETL backup page.

bstuder99:
I've gone ahead and added a listing to Summary.MD just in case, but I'd like to get some more feedback on this @srpomeroy

kaelanspatel:
> Is Storj significant enough on the topic of long term storage to justify its own article.
In my opinion probably not, actually, I just erred on the side of completeness. If we feel like it's redundant, I'm all for removing that section.

bstuder99:
Ok, let's leave it as part of the backup doc

bstuder99 (approved):

PR #557, opened 2023-02-21
first pass of docs for budget rules

nickcurie:
Planning to get this into v1.101

PR #569, opened 2023-02-27
Document /prediction/speccost API

michaelmdresser:


bstuder99:
Similar to this PR https://github.com/kubecost/docs/pull/557, I will plan to modify this doc through GitBook using advanced functionality. Few questions about this:
* Is this currently considered a beta feature?
* Is kubectl cost a requirement to utilize this API?
* Based on some stuff that came up regarding POST commands, are params still able to be added into the path, or do they need to be provided as a JSON object in the body. ie, is this acceptable: `POST http://<your-kubecost-address>/model/prediction/speccost?clusterId=...&defaultNamespace=...`

michaelmdresser:
> Is this currently considered a beta feature?

Yes.

> Is kubectl cost a requirement to utilize this API?

No, but it is currently the easiest. We may eventually have a frontend page for this, but I'm not sure where that stands right now.

> Based on some stuff that came up regarding POST commands, are params still able to be added into the path, or do they need to be provided as a JSON object in the body. ie, is this acceptable: `POST http://<your-kubecost-address>/model/prediction/speccost?clusterId=...&defaultNamespace=...`

This endpoint is a nonstandard usage of POST, in which there are separate query params and request body. Still valid, though!

michaelmdresser:
Just let me know what you need from me! I work a lot easier in Markdown which is why I submitted it like this. I'm perfectly happy if you want to port into GitBook specific format if that works better for your flow.

srpomeroy:
I have nothing to add to this PR besides my excitement for this new API. 😁

predict-spec-cost-api.md
```diff
@@ -0,0 +1,145 @@
...
+| `clusterID` | string (required) | The Kubecost cluster ID of the cluster in which the workloads will be deployed. Currently, this must be the same as the cluster ID of the Kubecost installation which is serving the `/speccost` endpoint. Support for multi-cluster is planned. |
+| `defaultNamespace` | string (required) | The namespace in which namespace-scoped objects will be "deployed" to if no namespace is set in the standard metadata field of the object. |
+| `window` | string (default `2d`) | The Kubecost data window used for determining resource costs fed into the cost prediction. |
+| `noUsage` | bool (default `false`) | Set to `true` to ignore historical usage data (if it exists) when making the prediction. This is equivalent to making a prediction using only requests. |
+
+The API requires that workloads be passed in the request body in YAML format and
+that the `Content-Type` header be set to `application/yaml`. Multiple workloads
+can be passed via separation with the standard `---` syntax.
+
+### Example
```
linhlam-kc, line 27: The example is nice. However, I think it would be easier for reader to follow if we break it into 2 parts:
- 1. create an example deployment manifest and pass it into env var
- 2. call  `/model/prediction/speccost` with curl to retrieve the cost prediction.

linhlam-kc (commented):
It is great to see this API is going to be available soon ! I just have a minor comment for the example. Another the thing you might want to add is to explain the example result and show how awesome it is to share to cost diff with current workload if there are any (from my understanding).
One quick question, does this API offer 1:1 features parity with `kubectl cost predict`?

predict-spec-cost-api.md
michaelmdresser: Yes, that would probably be better -- or even write it to a file, then pass to curl with the `@` syntax (or whatever it is...). I just copy-pasted my testing block here, which is why its all one big nasty shell script.

I should be able to make an update soon.

michaelmdresser:
> Another the thing you might want to add is to explain the example result and show how awesome it is to share to cost diff with current workload if there are any (from my understanding).

Great point, I definitely didn't highlight the diffing (or the incorporation of historical usage!) enough.

> One quick question, does this API offer 1:1 features parity with kubectl cost predict?

This API is what backs the upcoming version of `kubectl cost` (coinciding with v1.101 release of Kubecost). That's WIP, but you'll get a sneak preview if you'll be at tomorrow's prediction overview meeting.

In short, `kubectl cost predict` should have exact parity with this endpoint because it uses (will use this endpoint.

michaelmdresser:
@linhlam-kc made a few updates based on your feedback, let me know what you think!

linhlam-kc (approved):
@michaelmdresser Thank you ! It looks great to me now.

PR #574, opened 2023-03-03
Creates doc for Audit dashboard

bstuder99:
Creates Audit dashboard doc. Due to lack of interaction with the UI, there isn't a lot to document. Will have added hyperlinks to the Predict and Cost Events Audit APIs once both those pages go live. Content is currently slim but we can expand this over time as functionality is added.


nealormsbee (dismissed):
Looks good to me!

Issue #590, opened 2023-03-16
ETL initialism is not defined


MikeFindsThings:
The initialism ETL (extract, transform, load) is used many times throughout our documentation but never defined. [Several](https://owl.purdue.edu/owl/research_and_citation/apa_style/apa_formatting_and_style_guide/apa_abbreviations.html) [style](https://www.scientific-editing.info/blog/using-abbreviations-in-chicago-format/) [guides](https://guides.lib.uw.edu/hsl/ama/format) require spelling out initialisms the first time they are used, if not commonly known. Depending on the target audience for documentation (platform engineers/end users/FinOps/DevOps), ETL might not be commonly known.

MikeFindsThings:
Happy to change this is others agree :)

bstuder99:
Hi Mike. Great issue to bring up. In general, the policy in our style guide is also to spell acronyms long-hand the first time they appear in article before using the shorthand. However, there are certain acronyms that are exceptions because they are either universally understand (UI, OS), or because the meaning of the acronym is not as recognizable as the acronym itself (HTML). In the case of ETL, I think the direct long-hand extract, transform, load is a bit vague and does not necessarily enforce someone's understanding of the concept, nor is it a niche or product-specific one. However, I do agree that general understanding of how ETL pipelines work with Kubecost to provide cost metrics can be improved. For the time being, consult the [Kubecost Glossary](https://docs.kubecost.com/troubleshooting/kubecost-glossary) which explains terms like this.

bstuder99:
We've clarified the context of ETL in several of our docs in https://github.com/kubecost/docs/pull/592, thank you for pointing this out.

PR #592, opened 2023-03-17
Clarifies ETL acronym across several docs

bstuder99:
Based on https://github.com/kubecost/docs/issues/590, docs will elaborate slightly on ETL acronym by spelling longhand where it appears

### DIFF
--- a/etl-backup.md
+++ b/etl-backup.md
@@ -1,6 +1,6 @@
 # ETL Backup
 
-Kubecost's ETL is a computed cache based on Prometheus's metrics, from which the user can perform all possible Kubecost queries. The ETL data is stored in a `PersistentVolume` mounted to the `kubecost-cost-analyzer` pod.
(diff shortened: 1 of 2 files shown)


thomasvn:
This is great! Definitely think it's helpful to be more explicit about the ETL acronym.

One thing to note is that we often use the word ETL in two contexts:

1. The process of extract, transform, load. Which is what you often reference in your docs.
2. The cache/file that is generated as a result of the above process.

Just thought to mention this because it may affect the sentencing when we refer to ETL.

thomasvn (approved):

bstuder99:
I did not know about that distinction, good to know. I can add that to the Glossary as well. Just confirming, in the three articles above, ETL refers to the process itself than the cache/file, correct?

thomasvn:
@bstuder99 In `etl-backup.md` we're referring to the backup of a file to a cloud bucket. The others look good!

thomasvn (approved):

PR #631, opened 2023-04-28
Docs for Federated ETL's `useExistingS3Config`

thomasvn:
Add details about how to configure "useExistingS3Config" into a dropdown. Also add a disclaimer that this method is not recommended because misconfiguring it may have adverse effects.


rossfisherkc (approved):
LGTM

bstuder99:
Will think about how/should we capitalize Primary/Federated Cluster, but I don't think it's a current policy in our style guide. Will make edits later, I think this should be good to merge

PR #633, opened 2023-05-04
Prefix KUBECOST_PROJECT_ID with $

xbglowx:


thomasvn:
@xbglowx Thanks for that fix! Looks good to me.

thomasvn (approved):

PR #637, opened 2023-05-10
Update titles in GCP Cloud Integration doc for more readability

thomasvn:
This is just one proposed method to increase readability. Not sure if it's the best method.


thomasvn:
Alternatively, it could be "Option 3a" and "Option 3b"

bstuder99:
Totally understand the intent here, yes we should distinguish that these are exclusive options to achieve similar objectives.

bstuder99 (approved):

PR #639, opened 2023-05-15
Querying on-demand is not based off reconciled costs

thomasvn:
Add a note about this, as it can be a point of confusion for users.


bstuder99 (approved):

zorrobyte (commented):
LGTM

PR #643, opened 2023-05-16
enterprise upgrade compatibility

jessegoodier:
1st pass at advice on agent > primary compatibility. 

Need some help @bstuder99 especially with formatting and nav bar changes.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -32,6 +32,7 @@
       * [Azure Long Term Storage](long-term-storage-azure.md)
     * [GCP](install-and-configure/install/long-term-storage/gcp/README.md)
       * [GCP Long Term Storage](long-term-storage-gcp.md)
+    * [Enterprise Upgrading/Compatibility Matrix](upgrading.md)
   * [ETL Backup](etl-backup.md)
     * [Sharing ETL Backups](sharing-etl-backups.md)
   * [Provider Installations](install-and-configure/install/provider-installations/README.md)
--- a/upgrading.md
+++ b/upgrading.md
@@ -1,0 +1,34 @@
+# Enterprise Upgrading / Compatibility Matrix
+
+## Overview
+
+Kubecost is flexible with version differences between Primary and Agent clusters.
+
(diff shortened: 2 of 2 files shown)


jessegoodier:
> Couple notes. Have already incorporated most of my feedback into docs edits:
> 
> * Are agent clusters and secondary clusters synonymous? Want to make sure the language we use is precise. yes, exactly.
> * This looks like a new doc? We need to still update summary.md. Would this belong in the Troubleshooting section, or Advanced Configuration? I'll leave nav to others. probably link from here though https://docs.kubecost.com/install-and-configure/install/multi-cluster
> * Is this doc going to have to be updated routinely over time as new Kubecost versions come out? ideally with the release
> * Can we link to release notes also? for sure

bstuder99:
Ok, summary.md has been updated and the release notes link has been added. I personally don't like the idea of a living doc that needs monthly updates, I think it might be better to provide the oldest recommended agent version, some version ideally in the last few months which can be updated more casually.

thomasvn:
@jessegoodier How is testing performed for the multi-cluster compatibility matrix? Is this currently done manually with differing primary & agent versions?

jessegoodier:
To my knowledge, there hasn't been a breaking change between primary and secondary enterprise versions in years/ever.

IMO: It would come down to things that we are aware of like prom label (pod_name) or etl format changes that would impact single cluster upgrades anyway. 

Do you have other concerns?

One of the goal of this doc is to help customers understand the low risk of version upgrades and version drift.

thomasvn:
I agree that version drift between primary and agent (secondary) is _generally_ low/medium risk. But I'm hesitant to be telling users that this is an officially supported state of running Kubecost.

Small changes to the Helm chart, the Prometheus subchart (e.g. metric collection & metric relabeling), or the backend's ETL pipeline (e.g. introduction of weekly ETLs) can have an impact on the user's deployment.

If we do plan to roll out this doc, I think we should clearly communicate that there is some risk in having version drift between primary and agent (secondary). The biggest risk being data integrity.

PR #651, opened 2023-05-25
aws-cur-best-practice

jessegoodier:


mmurph3 (commented):
@jessegoodier , Chris M. brought up a good point. Could we add more context here to make it more clear on what config files need to be downloaded and to whether it applies to account where kubecost primary lives or master payer? https://github.com/kubecost/docs/pull/651/commits/8e42f1d89b785d5fa84506608c3376836a34ad69#diff-fb0c175b18f0ce22792df1c597d32896179f566431ee6da902d1cb7b51401415R117

jessegoodier:
Yes! I think that's the last step. will try to get this out soon. Feel free to co-author as well @mmurph3 and Chris M.

jessegoodier:
@mmurph3 
added link to policy docs. I believe this is good to go.

jessegoodier:
@bstuder99 this guide is meant to be our best practice for setting up the CUR with Kubecost. I think some links from https://docs.kubecost.com/install-and-configure/install/cloud-integration/aws-cloud-integrations

and the navbar would be useful. Do you mind taking lead on that?

bstuder99:
Little curious about the user journey involved in using this doc. This is still considered a complete AWS integration, just as part of a multi-cloud setup, correct? Is the idea that we should direct users here instead of our regular AWS integration guide as long as their environment meets the three stated conditions at the top of the article?

jessegoodier:
> Little curious about the user journey involved in using this doc. This is still considered a complete AWS integration, just as part of a multi-cloud setup, correct? Is the idea that we should direct users here instead of our regular AWS integration guide as long as their environment meets the three stated conditions at the top of the article?

You got it Brett. See latest commit for potential wording.

bstuder99:
Ok, I've gone through with a few edits. Everything looks good

bstuder99 (dismissed):

jessegoodier:
@bstuder99 had to merge in main conflicts

bstuder99 (approved):

chipzoller (commented):
Based on the image refactoring proposed in #698, I would suggest a couple things:

1. Make image captions descriptive of the image itself because they also function as alt text.
2. Begin the path reference with a forward slash. I found that GitBook will not interpret this correctly if it is later indented in a structured block. Standardizing on leading forward slash is also fairly common for projects which use a root-level directory for images and assets.

jessegoodier:
> Based on the image refactoring proposed in #698, I would suggest a couple things:
> 
> 1. Make image captions descriptive of the image itself because they also function as alt text.
> 2. Begin the path reference with a forward slash. I found that GitBook will not interpret this correctly if it is later indented in a structured block. Standardizing on leading forward slash is also fairly common for projects which use a root-level directory for images and assets.

100% feel free to start a new PR on this. @bstuder99, see above. also feel free to help out if you have time.

PR #658, opened 2023-06-01
Remove storage.md from file tree

thomasvn:
This document was pertinent back in Kubecost ~v1.67, but likely no longer applies to a large portion of existing users


bstuder99:
So it looks like we'd be removing the article from public display, but the raw MD file will still exist in the docs repo. That's fine, but have you confirmed with anyone else if this removal is fine? GitBook is showing be abt 2000 page views on this article, so I want to make sure it's not critical or even helpful if we purge this.

thomasvn:
Haven't spoken to the team about this yet. If you're seeing significant page views here, we can leave this doc as is!

bstuder99:
We can table this for now, but if we want to make updates to this doc or eventually delist it, I'm down. Just think this may require a little more consultation first

PR #659, opened 2023-06-06
Configuring IAM for Spot Data Feed

thomasvn:
This information exists in different places throughout the docs & helm chart & poc-common-configs. I thought to consolidate here.

Related to: https://github.com/opencost/opencost/pull/1966

### DIFF
--- a/aws-spot-instances.md
+++ b/aws-spot-instances.md
@@ -1,12 +1,23 @@
 # AWS Spot Instances
 
-## Spot Data feed integration
+## Considerations before configuring spot pricing
 
-Kubecost will reconcile your Spot prices with Cost and Usage Reports (CURs) as they become available (usually 1-2 days), but pricing data can be pulled hourly by integrating directly with the AWS Spot feed. AWS CUR integration and Spot data feed are two methods for Kubecost to read your billing data. The CUR is updated daily, while the Spot data feed is updated hourly by only applies to Spot Instances. Reduce complexity by only integrating with one of these methods. To enable Spot data feed, follow AWS' [Spot Instance data feed](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-data-feeds.html) doc.
+Kubecost uses onDemand pricing from the cloud providers until the actual cloud bill is available. This is almost always ready in 48 hours. Most users will likely prefer to configure [AWS cloud-integrations](aws-cloud-integrations.md) and skip the below setup.
+
+Kubecost will reconcile all onDemand resource prices when Cost and Usage Reports (CURs) are available (usually 1-2 days)
+
+For users that have a majority of their costs from spot nodes, the below guide can increase short-term (<48 hour) node costs. Note that all other costs will still be onDemand, which is why the below guide should be considered optional.
+
 
 ## Configuring the Spot data feed in Kubecost
 
-These values can either be set from the Kubecost UI or via `.Values.kubecostProductConfigs` in the Helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be deleted on pod restart.
+With Kubecost spot pricing- data can be pulled hourly by integrating directly with the AWS Spot feed.
+
+First- to enable the AWS Spot data feed, follow AWS' [Spot Instance data feed](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-data-feeds.html) doc.
+
+When configuring note the settings used as these values will be needed for the Kubecost configuration.
+
+There are multiple options- this can either be set from the Kubecost UI or via `.Values.kubecostProductConfigs` in the Helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be deleted on pod restart.
 
 * `projectID` the Account ID of the AWS Account on which the Spot nodes are running.
 * `awsSpotDataRegion` region of your Spot data bucket
--- a/aws-spot-instances.md
+++ b/aws-spot-instances.md
@@ -2,7 +2,7 @@
 
 ## Considerations before configuring spot pricing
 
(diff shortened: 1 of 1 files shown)


jessegoodier:
I added some stronger text around using cloud-integration instead. Happy to revert if you prefer your original.

thomasvn:
@jessegoodier I'm good with the change! I agree that cloud integration should cover the majority of use cases.

srpomeroy:
Don't we use "generic" spot pricing if the spot label is on the node and the spot data feed is not hooked up?
image

thomasvn:
@srpomeroy Yes I think so! https://github.com/opencost/opencost/blob/v1.104/pkg/cloud/aws/provider.go#L808-L810

aws-spot-instances.md
```diff
@@ -1,12 +1,20 @@
 # AWS Spot Instances
 
-## Spot Data feed integration
+## Considerations before configuring spot pricing
 
-Kubecost will reconcile your Spot prices with Cost and Usage Reports (CURs) as they become available (usually 1-2 days), but pricing data can be pulled hourly by integrating directly with the AWS Spot feed. AWS CUR integration and Spot data feed are two methods for Kubecost to read your billing data. The CUR is updated daily, while the Spot data feed is updated hourly by only applies to Spot Instances. Reduce complexity by only integrating with one of these methods. To enable Spot data feed, follow AWS' [Spot Instance data feed](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-data-feeds.html) doc.
+Kubecost uses onDemand pricing from the cloud providers until the actual cloud bill is available. This is almost always ready in 48 hours. Most users will likely prefer to configure [AWS cloud-integrations](aws-cloud-integrations.md) and skip the below setup.
```
srpomeroy, line 5:
```suggestion
Kubecost uses public pricing from the cloud providers until the actual cloud bill is available. This is almost always ready in 48 hours. Most users will likely prefer to configure [AWS cloud-integrations](aws-cloud-integrations.md) and skip the below setup.
```

aws-spot-instances.md
```diff
@@ -1,12 +1,20 @@
...
+
+For users that have a majority of their costs from spot nodes, the below guide can increase short-term (<48 hour) node costs. Note that all other costs will still be onDemand, which is why the below guide should be considered optional.
```
srpomeroy, line 7:
```suggestion
For users with most of their costs from spot nodes, the guide below can increase short-term (<48 hour) node costs. Note that all other costs will still be based on public pricing, which is why the below guide should be considered optional.
```

aws-spot-instances.md
```diff
@@ -1,12 +1,20 @@
...
 ## Configuring the Spot data feed in Kubecost
 
-These values can either be set from the Kubecost UI or via `.Values.kubecostProductConfigs` in the Helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be deleted on pod restart.
+With Kubecost spot pricing- data can be pulled hourly by integrating directly with the AWS Spot feed.
+
+First- to enable the AWS Spot data feed, follow AWS' [Spot Instance data feed](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-data-feeds.html) doc.
+
+When configuring note the settings used as these values will be needed for the Kubecost configuration.
+
+There are multiple options- this can either be set from the Kubecost UI or via `.Values.kubecostProductConfigs` in the Helm chart. Note that if you set any kubecostProductConfigs from the Helm chart, all changes via the frontend will be deleted on pod restart.
```
srpomeroy, line 17:
```suggestion
There are multiple options- this can either be set from the Kubecost UI or via `.Values.kubecostProductConfigs` in the Helm chart. Note that if you set any `kubecostProductConfigs` from the Helm chart, all changes via the frontend will be deleted on pod restart.
```

aws-spot-instances.md
```diff
@@ -23,6 +31,69 @@ Spot data feeds are an account level setting, not a payer level. Every AWS Accou
...
+Kubecost requires read access to the Spot data feed bucket. The following IAM policy can be used to grant Kubecost read access to the Spot data feed bucket.
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+        "Sid": "SpotDataFeed",
+        "Effect": "Allow",
+        "Action": [
```
srpomeroy, line 45: Suggest revising permissions to the following:

```
"s3:GetBucketAcl",
"s3:GetBucketLocation",
"s3:GetObject",
"s3:ListBucket"
```

srpomeroy (changes requested):
Some nits based on ondemand vs public pricing as well as IAM permissions.

aws-spot-instances.md
```diff
@@ -23,6 +31,69 @@ Spot data feeds are an account level setting, not a payer level. Every AWS Accou
...
+  ]
+}
+```
+
+To attach the IAM policy to the Kubecost service account, you can use IRSA or the account's service key.
+
+### Option 1: IRSA (IAM Roles for Service Accounts)
+
+```bash
+eksctl create iamserviceaccount \
```
srpomeroy, line 64: Will this overwrite an existing service account annotation?

aws-spot-instances.md
thomasvn: @srpomeroy I used `policy/kc-spot-feed-us-east-2-policy` as reference for this doc, and I also used that in my repro. Have you tested the above permissions? If so I can make the change, as your suggestion would be more restrictive!

aws-spot-instances.md
thomasvn: Hmm good catch, I think it probably will. Would it help to add the following note?

> If your `serviceaccount/kubecost-cost-analyzer` already has IRSA annotations attached, be sure to include all policies necessary when running this command.

srpomeroy: That works!

aws-spot-instances.md
```diff
@@ -1,12 +1,20 @@
 # AWS Spot Instances
 
-## Spot Data feed integration
+## Considerations before configuring spot pricing
```
bstuder99, line 3: *Spot

srpomeroy (approved):
LGTM

PR #661, opened 2023-06-07
Update advanced-reports.md

srpomeroy:


advanced-reports.md
```diff
@@ -1,4 +1,4 @@
-# Advanced Reporting
+# Advanced Reporting - Beta
```
srpomeroy, line 1: THis should be ALpha

srpomeroy, line 1:
```suggestion
# Advanced Reporting - Alpha
```

PR #667, opened 2023-06-09
Manually updates Allocation API doc to correct unintentional indents

bstuder99:
Based on an existing docs issue, I am manually reverting double indents in this doc which seems to be a symptom of GitBook being unable to register changes when syncing with GitHub. Seeking a more permanent resolution from GB.

### DIFF
--- a/assets-api.md
+++ b/assets-api.md
@@ -18,7 +18,7 @@ When set to `false`, this endpoint returns daily time series data vs cumulative 
 {% endswagger-parameter %}
 
 {% swagger-parameter in="path" name="disableAdjustments" type="boolean" required="false" %}
-When set to `true`, zeros out all adjustments from cloud provider reconciliation, which would otherwise change the totalCost. Default value is`false`.
+When set to `true`, zeros out all adjustments from cloud provider reconciliation, which would otherwise change the totalCost. Default value is `false`.
 {% endswagger-parameter %}
 
 {% swagger-parameter in="path" name="format" type="string" required="false" %}
--- a/assets-api.md
+++ b/assets-api.md
@@ -38,7 +38,7 @@ Filter results by cluster ID, which is generated from `cluster_id` provided duri
 {% endswagger-parameter %}
 
 {% swagger-parameter in="path" name="filterLabels" type="string" required="false" %}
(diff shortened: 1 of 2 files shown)


assets-api.md
```diff
@@ -14,113 +14,55 @@ Used to consolidate cost model data. Supported aggregation types are cluster and
...
-
-.
+When set to `false`, this endpoint returns daily time series data vs cumulative data. Default value is `false`.
 {% endswagger-parameter %}
 
 {% swagger-parameter in="path" name="disableAdjustments" type="boolean" required="false" %}
-When set to `true`, zeros out all adjustments from cloud provider reconciliation, which would otherwise change the totalCost. Default value is
-
-`false`.
+When set to `true`, zeros out all adjustments from cloud provider reconciliation, which would otherwise change the totalCost. Default value is`false`.
```
thomasvn, line 21:
```suggestion
When set to `true`, zeros out all adjustments from cloud provider reconciliation, which would otherwise change the totalCost. Default value is `false`.
```

assets-api.md
```diff
@@ -14,113 +14,55 @@ Used to consolidate cost model data. Supported aggregation types are cluster and
...
-Filter results by cloud label or cloud tag. For example, appending
-
-`&labels=deployment:kubecost-cost-analyzer`
-
-only returns assets with label
-
-`deployment=kubecost-cost-analyzer`. Note that subparameter `:` symbols are required to denote
-
-`<labelKey>:<labelValue>` pairs.
+Filter results by cloud label or cloud tag. For example, appending `&labels=deployment:kubecost-cost-analyzer` only returns assets with label`deployment=kubecost-cost-analyzer`. Note that subparameter `:` symbols are required to denote `<labelKey>:<labelValue>` pairs.
```
thomasvn, line 41:
```suggestion
Filter results by cloud label or cloud tag. For example, appending `&labels=deployment:kubecost-cost-analyzer` only returns assets with label `deployment=kubecost-cost-analyzer`. Note that subparameter `:` symbols are required to denote `<labelKey>:<labelValue>` pairs.
```

bstuder99:
Thanks for catching those, awaiting review

thomasvn (approved):

bstuder99:
Ok it looks GitBook doesn't like that. I will have to find a workaround

PR #673, opened 2023-06-15
Add v2.1 filter documentation

michaelmdresser:


filters-api.md
```diff
@@ -19,38 +23,80 @@ The supported filter fields as of v1.96 are:
...
-
-Each individual filter is separated by a `+`, representing logical AND.
+Added in v1.105 of Kubecost:
+* `department`
+* `environment`
+* `owner`
+* `product`
+* `team`
+
+The supported filter ops in v1.96 of Kubecost are:
```
srpomeroy, line 33: suggest replacing `ops` with `operators`
```suggestion
The supported filter operators in v1.96 of Kubecost are:
```

filters-api.md
```diff
@@ -19,38 +23,80 @@ The supported filter fields as of v1.96 are:
...
+  * For slice/array fields (e.g. services): slice contains at least one value equal (equivalent to `~:`)
+  * For map fields (e.g. labels): map contains key (equivalent to `~:`)
+* `!:` Inequality, or "not contains" if an array type
+
+Added in v1.105 of Kubecost:
+* `~:` Contains
+  * For string fields: contains
+  * For slice fields: slice contains at least one value equal (equivalent to `:`)
+  * For map fields: map contains key (equivalent to `:`)
+* `!~:` NotContains, inverse of `~:`
```
srpomeroy, line 45: `NotContains` reads like a property or command flag. Would it make more sense to split these words?
i.e.: not contains or does not contain

srpomeroy (dismissed):
Very exciting to see this type of filter logic coming to Kubecost!
a few nits but nothing earth-shattering. 

I assume the examples and logic are accurate but I have not verified.

filters-api.md
michaelmdresser: Great suggestion, thank you

filters-api.md
michaelmdresser: I lean towards keeping it as `NotContains` because if we ever get into a conversation about operators using anything other than the operator itself (`!~:` in this case), I'd rather see consistent (if esoteric) naming to smooth the likely-complex conversation.

I'm still open to changing it if you think it's detrimental to the effectiveness of the documentation, though! The above is an opinion I'm totally willing to relinquish.

bstuder99 (approved):

PR #679, opened 2023-06-23
Remove statement about deletion of CUR files

thomasvn:


srpomeroy (approved):

bstuder99:
Is there a reason we are no longer detailing this? I think this comes from an older doc of misc. CSP best practices, and I'm just curious if this info is inaccurate

bstuder99:
Fine to merge this, just hoping for more context. We have a few other examples like this in some of our cloud billing int. docs.

srpomeroy:
> Is there a reason we are no longer detailing this? I think this comes from an older doc of misc. CSP best practices, and I'm just curious if this info is inaccurate

AWS does not publish an official recommendation for lifecycle policies on CUR data. Plus, we use Athena to query the CUR data and are still determining any repercussions from deleting data that Athena assumes to have access to.

PR #686, opened 2023-07-17
xlink cloud metrics and clean up wording

jessegoodier:
The list of tags is coming up frequently- adding link from cluster detail page.

The cloud-cost-metrics is a bit wordy, this is a first attempt to simpilfy.


jessegoodier:
@bstuder99 can you verify the `\` usage here? I'm not certain what is/isn't needed.

also feel free to reword, looking to simplify where possible.

bstuder99 (dismissed):
Thanks for cleaning this up

PR #687, opened 2023-07-18
Update implementation link in filters-api.md

rayhatfield:
Noticed that the existing lexer/parser link [in the filters documentation](https://docs.kubecost.com/apis/apis-overview/filters-api) 404s. Updated to point to `/pkg/filter21/allocation` (which I'm not 100% sure is the correct place).

### DIFF
--- a/filters-api.md
+++ b/filters-api.md
@@ -113,4 +113,4 @@ curl -G 'localhost:9090/model/assets' \
 
 ## Formal grammar and implementation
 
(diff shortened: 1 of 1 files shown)


michaelmdresser:
Nearly right! The correct link there is https://github.com/opencost/opencost/tree/develop/pkg/filter21/ast. Thanks for the PR @rayhatfield!

bstuder99 (approved):

Issue #691, opened 2023-07-26
Non-root path ingress example needs updating to current Ingress versions


omkensey:
The Ingress [Non-root path example](https://docs.kubecost.com/install-and-configure/install/ingress-examples#non-root-path-example) still uses the v1beta1 API group; it should be updated for the v1 networking.k8s.io API like the other Ingress example resources on the same page.  (I'll submit a PR for this if no one gets to it first in the next couple of days.)

bstuder99:
Thank you for pointing this out. The issue has been corrected.

PR #692, opened 2023-07-27
Update YAML to match the Ingress v1 schema

omkensey:
Followup from kubecost/docs#691


thomasvn (approved):
@omkensey Thanks for the fix! Looks good to me.

cc @bstuder99 @jessegoodier

PR #693, opened 2023-07-28
Edit instructions for v1.104 ETL repair

thomasvn:
- Removed instruction to delete files from `/combined` dir. Just running a repair on the primary's cluster _should_ be enough to trigger the federator to overwrite any corrupted files in the `/combined` dir. NOTE: I'm still in the process of validating whether this is true.
- Provide additional comments on how to validate success


nikovacevic (approved):

thomasvn:
@jessegoodier Testing showed that you don't need to delete files in the `/combined` dir! Also, I think leaving this `/model/etl/asset/repair` as a standalone is ok because this same doc has detailed instructions on how to ETL repair.

Ready to merge!

PR #730, opened 2023-09-20
Update ETL repairs doc from CloudUsage to CloudCost

thomasvn:
## Proposed Changes

Update ETL repairs doc from CloudUsage to CloudCost


bstuder99 (approved):

srpomeroy:
@thomasvn Should we not save the legacy process for those not on a version where etlCloudUsage is disabled and cloudCosts are enabled?

thomasvn:
@srpomeroy Good callout! I'll author a new PR which includes both the new CloudCost repair and also mentions the old CloudUsage repair.

PR #733, opened 2023-09-21
Include instructions for CloudUsage ETL repair for users still running it

thomasvn:
## Related issue

Amends PR https://github.com/kubecost/docs/pull/730

## Proposed Changes

Include instructions for CloudUsage ETL repair for users still running it

### DIFF
--- a/etl-repair.md
+++ b/etl-repair.md
@@ -69,7 +69,7 @@ $ kubectl logs deploy/kubecost-cost-analyzer | grep CloudCost
 ```
 {% endcode %}
 
-By default, all users v1.106+ no longer have CloudUsage ETL enabled (`.Values.kubecostModel.etlCloudUsage`, `.Values.kubecostModel.etlCloudAsset`). If you are using CloudUsage ETL, use the commands below:
+By default, Kubecost v1.106+ no longer has CloudUsage ETL enabled (`.Values.kubecostModel.etlCloudUsage`, `.Values.kubecostModel.etlCloudAsset`). If you are using CloudUsage ETL, use the commands below:
 
 {% code overflow="wrap" %}
 ```bash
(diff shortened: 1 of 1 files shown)


etl-repair.md
```diff
@@ -69,6 +69,19 @@ $ kubectl logs deploy/kubecost-cost-analyzer | grep CloudCost
 ```
 {% endcode %}
 
+By default, all users v1.106+ no longer have [CloudUsage ETL](https://github.com/kubecost/cost-analyzer-helm-chart/blob/63dea06ae1ea771a8622e786b22f6f6625a49349/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L788-L794) enabled. If you are using CloudUsage ETL, use the commands below:
```
chipzoller, line 72: I really dislike mentioning specific line numbers in any code in the documentation because it almost immediately gets outdated. I instead like to mention which value or, just generally, which functionality may need to be enabled and point to the file and let users find the exact location.

thomasvn: @chipzoller Removed the link! Replaced with helm config value.

etl-repair.md
```diff
@@ -69,6 +69,19 @@ $ kubectl logs deploy/kubecost-cost-analyzer | grep CloudCost
 ```
 {% endcode %}
 
+By default, all users v1.106+ no longer have CloudUsage ETL enabled (`.Values.kubecostModel.etlCloudUsage`, `.Values.kubecostModel.etlCloudAsset`). If you are using CloudUsage ETL, use the commands below:
```
srpomeroy, line 72: The wording feels a bit off after the comma. 

How about something like: 
```suggestion
By default, Kubecost v1.106+ no longer has CloudUsage ETL enabled (`.Values.kubecostModel.etlCloudUsage`, `.Values.kubecostModel.etlCloudAsset`). If you are using CloudUsage ETL, use the commands below:
```

srpomeroy (changes requested):

etl-repair.md
thomasvn: I like it!

chipzoller (approved):

PR #740, opened 2023-09-29
add information on new oidc helm value

nickcurie:
## Related issue #

https://kubecost.atlassian.net/browse/SELFHOST-304

## Proposed Changes
Added information about a new OIDC helm value `skipOnlineTokenValidation`

### DIFF
--- a/user-management-oidc.md
+++ b/user-management-oidc.md
@@ -76,7 +76,7 @@ The token is then validated remotely in one of two ways:
 If `skipOnlineTokenValidation` is set to true, Kubecost will skip accessing the OIDC introspection endpoint for online token validation and will instead attempt to locally validate the JWT claims.
 
 {% hint style="danger" %}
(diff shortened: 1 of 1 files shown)


srpomeroy:
@nickcurie Will using this parameter also fix the Auth0 issues noted in the doc?

nickcurie:
At a glance I would say yes, though I think @kaelanspatel might be better equipped to answer that question. Presumably if an auth0 user has the skipOnlineValidation enabled then we'd just verify the token locally so introspection is not required

srpomeroy (approved):
LGTM. Is this available in a current release or slated for 1.107?

nickcurie:
This will be 1.107

bstuder99 (approved):

nickcurie:
Oops sorry guys looks like this is meant to be in 1.106. I am cherry-picking the changes now

PR #745, opened 2023-10-05
Updates Install Troubleshooting guide with additional example

bstuder99:
## Proposed Changes

Just wanted to add an example from my personal experience installing Kubecost, where re-installing Kubecost may result in a failure to grab the Helm chart. Not sure how universal this problem is or how commonly-known the solution should be, as it's very simple. Can someone maybe read this for accuracy? My specific error message is:

```
Error: INSTALLATION FAILED: failed to download "https://kubecost.github.io/cost-analyzer/cost-analyzer-1.106.2.tgz"
```

### DIFF
--- a/troubleshoot-install.md
+++ b/troubleshoot-install.md
@@ -4,7 +4,7 @@ Once an installation is complete, access the Kubecost UI to view the status of t
 
 ## General troubleshooting commands
 
-These Kubernetes commands can be helpful when finding issues with deployments:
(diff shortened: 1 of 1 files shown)


chipzoller (commented):
I don't personally feel this is that valuable of a troubleshooting step as it kind of states the obvious. If the chart can't be downloaded it's going to be something on the user's side.

jessegoodier (approved):

PR #747, opened 2023-10-09
Creates Persistent Volume Right-Sizing Recommendation doc

bstuder99:
## Proposed Changes

Creates doc for Savings Insight PV Right-Sizing Recs. Definitely need details on recommendation calculation verified here, will ask around for a core engineer to review


using-kubecost/navigating-the-kubecost-ui/savings/pv-right-sizing-rec.md
```diff
@@ -0,0 +1,19 @@
...
+
+Kubecost is able to provide recommendations for resizing your PVs by comparing their average usage to their maximum capacity, and can recommend sizing down to smaller storage sizes.
+
+To access the Persistent Volume Right-Sizing Recommendations page, select *Savings* from the left navigation, then select *Right-size persistent volumes*.
+
+![Table](/images/rightsizingpv.png)
+
+Kubecost will display a table containing all PVs in your environment. Table columns include the PV name and its corresponding cluster, and metrics pertaining usage and savings. The estimated savings per month per table item is calculated by subtracting your recommended cost from the current cost.
+
+Recommendations appear in multiple denominations of capacity, rounded up. The smallest denomination Kubecost will recommend per PV is 1.1 GB. From here, the recommended capacity increases in intervals of 1 GB.
```
nickcurie, line 11: This is more of a front end display issue but I still want to point it out. We are actually calculating disk usage and disk size recommendations in GiB (1024 * 1024 * 1024), but displaying them in the table in GB (1000 * 1000 * 1000) which explains the extra 0.1 GB. While it might appear that we only increment by 1 GB, it would be more accurate to say we increment by 1 GiB.

using-kubecost/navigating-the-kubecost-ui/savings/pv-right-sizing-rec.md
```diff
@@ -0,0 +1,19 @@
...
+
+You can filter your table of PVs using the Cluster dropdown to view PVs in an individual cluster, or across all connected clusters.
+
+You can also adjust Kubecost’s average recommended capacity size using the Profile dropdown, which establishes how much minimum excess capacity you will for every PV, using their local usage data from the past six hours. The percentage value associated with each Profile is the minimum unused capacity required per PV, which is then added to the max usage to obtain Kubecost’s recommendation. Recommended capacity is calculated as (max usage + (max usage * overhead percentage)), then rounded up to the nearest capacity increment. The higher the minimum excess capacity needed, the higher the average recommended capacity, and therefore the lower the average savings.
+
+For example, for a PV with a max usage of 2 GB, and a selected Production Profile (which requires 50% excess capacity), the overhead will be calculated as 2 * .5, then added to the max usage, resulting in a minimum recommended capacity of 3 GB. This should then be rounded up to 3.1 GB for the final recommendation.
```
nickcurie, line 17: See above comment for why this is reading weird.

nickcurie (commented):
Left some comments that should hopefully clear up some of our "strange" reasoning from a user perspective w/r/t our +0.1 GB recommendations. Its ultimately a code issue not a docs issue, but I still wanted to point it out in case you wanted to alter the wording.

using-kubecost/navigating-the-kubecost-ui/savings/pv-right-sizing-rec.md
bstuder99: That makes sense, thanks for explaining that. I'll clarify that.

using-kubecost/navigating-the-kubecost-ui/savings/pv-right-sizing-rec.md
bstuder99: Also, we display Max Usage in GB as well, but is this also converted from GiB? For example, when Max Usage is listed as 1.1 GB, because there's a slight uptick in capacity from the conversion, can we assume the Max Usage in GiB is close to ~1?

nickcurie (approved):

PR #750, opened 2023-10-10
Update custom-prom.md

yeowchyuan:


thomasvn:
Thanks for this fix @yeowchyuan! This looks like the correct format for k8s DNS. An example of which found here: https://github.com/kubecost/cost-analyzer-helm-chart/blob/71b5e648a982a0da22bea0f34c26ce048a785f3e/cost-analyzer/values.yaml#L61

thomasvn (approved):

PR #753, opened 2023-10-11
Remove .gitbook/assets unused files

jpreynat:
## Related issue #

GitBook created a bunch of unused duplicate of files due to a bug (since fixed) during the Git sync process.

## Proposed Changes

This PR removes all now unused assets in this repository across the files to ensure they are also properly deleted from the GitBook side and won't be duplicated anymore.


bstuder99:
Hey Johan, thank you for helping clean up these unneeded files. I've checked these files to confirm they don't appear in any live docs and it looks good on my part. In the future, if we want to keep all image files within one folder group with descript file names (avoiding "image.png" instances), do you have a recommendation or best practices? Would it be better to implement images manually through GitHub?

bstuder99 (approved):

jpreynat:
Hey @bstuder99 👋 
Sorry, I meant to reply earlier but didn't find the time.

It's true that before, we used to name all images `image (n).png`, which was very difficult to handle on the GitHub side.
But if I'm not wrong, recently the images that you upload to GitBook should use the file names directly.
If you can confirm that it's not the case, then I'll have a better look at the reason why your new images wouldn't use the original name.

bstuder99:
Ok, I will keep that in mind the next time I upload an image.

PR #755, opened 2023-10-11
add useMultiClusterDB

jessegoodier:
Add `--set federatedETL.federator.useMultiClusterDB=true`
to GMP doc

### DIFF
--- a/gcp-gmp-integration.md
+++ b/gcp-gmp-integration.md
@@ -38,7 +38,8 @@ helm upgrade -i kubecost cost-analyzer/ \
 --set global.gmp.enabled="true" \
 --set global.gmp.gmpProxy.projectId="${GCP_PROJECT_ID}" \
 --set prometheus.server.global.external_labels.cluster_id="${CLUSTER_NAME}" \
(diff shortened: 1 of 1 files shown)


srpomeroy (approved):

thomasvn:
@jessegoodier @srpomeroy Thanks for the update! Although, this doc may need to be revisited in the future because there aren't any instructions yet on how to set up primaries & agents with a Federated ETL Architecture and GMP.

Therefore setting `federatedETL.federator.useMultiClusterDB=true` may have no effect. Oh, also the values is now `.Values.federatedETL.useMultiClusterDB`, so that may need updating ([code ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/71b5e648a982a0da22bea0f34c26ce048a785f3e/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L782-L785)).

srpomeroy:
As always, @thomasvn, thank you for your on-point feedback! I've opened a PR to adjust the value and an issue to capture the enhancements to the doc you suggested.

PR #759, opened 2023-10-11
Creates Admission Controller documentation

### LINKED ISSUE #749: [Enhancement] No documentation currently exists for Kubecost's Admission Controller feature
### Description

Specifically, the feature that this Helm config enables:

```yaml
kubecostAdmissionController:
  enabled: true
```

Note, that this was released as an experimental/beta feature. Its rough idea was to allow Kubecost admins to set budgets for certain deployments/namespaces. If you were to exceed the budget, the admission controller would not let you create the newly requested resource.

To get more info about the feature, we may need to talk to the engineers that wrote it. @bstuder99 Also your call as to whether it is a priority to further document this feature.

bstuder99:
## Related issue https://github.com/kubecost/docs/issues/749

Closes #749

### DIFF
--- a/using-kubecost/admission-controller.md
+++ b/using-kubecost/admission-controller.md
@@ -6,7 +6,7 @@ The Admission Controller is a beta feature. Please read the documentation carefu
 
 Kubecost's Admission Controller is a tool which leverages Kubecost's Predict API to display expected future costs for changes made to CPU and RAM in your workloads. This can be used to anticipate future spend before resources have been allocated.
 
-The Admission Controller exists in Kubecost's cost analyzer pod and uses a `ValidatingAdmissionWebhook` to send back cost data on every update you make to deployments in your Kubernetes clusters. The Predict API then determines the differences in cost, and provides that information in a convenient table in your terminal.
+The Admission Controller exists in Kubecost's cost-analyzer pod and uses a `ValidatingAdmissionWebhook` to send back cost data on every update you make to deployments in your Kubernetes clusters. The Predict API then determines the differences in cost, and provides that information in a convenient table in your terminal. Non-`kubectl` interactions will still be able to receive cost prediction data.
 
 ## Installing the Admission Controller
 
@@ -16,16 +16,21 @@ Before installing the Admission Controller, make sure you have [`kubectl`](https
 
 ### Installation
 
-The following command will enable the Admission Controller in your existing Kubecost application:
-
+1. This command sets up the TLS key and certificate for the Admission Controller.
 ```
-helm upgrade --install kubecost \
+$ git clone https://github.com/kubecost/cost-analyzer-helm-chart.git
+$ cd cost-analyzer-helm-chart/cost-analyzer
+$ kubectl create namespace kubecost
+$ ./scripts/create-admission-controller-tls.sh kubecost
+```
+
+2. This command will update your existing Kubecost application by enabling the Admission Controller.
+```
+$ helm upgrade --install kubecost \
   --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
-  --namespace kubecost --create-namespace \
+  --namespace kubecost \
   --set kubecostAdmissionController.enabled=true \
-git clone https://github.com/kubecost/cost-analyzer-helm-chart.git \
-cd cost-analyzer-helm-chart/cost-analyzer/scripts \
-./create-admission-controller-tls.sh <namespace-to-install-kubecost-in>
+  -f values.yaml
 ```
 
 You may need to wait several minutes for the controller to activate. You can check the status of the Admission Controller with `kubectl get service -n kubecost`, when the Admission Controller has been installed in the default `kubecost` namespace. Look for `webhook-server` to confirm a successfull install.
@@ -34,4 +39,4 @@ You may need to wait several minutes for the controller to activate. You can che
 
 When a Deployment is updated, the Kubernetes API will send requests containing Deployment information to the Kubecost pod, which will then be read for number of replicas, CPU requests, and RAM requests to calculate a monthly cost estimate. That estimate will be reported back to the client making the update. If the Kubecost pod is unable to respond to this request, the deployment will be added to the cluster without any information being sent back.
 
-To validate, you can run `kubectl edit deployment -n kubecost`, which should open a text file containing your deployment. Modify the CPU or RAM requests as desired, then save the file. In your terminal, your output will display CPU/RAM unit hours, total cost, and the cost difference (`diff` column).
+To validate, you can run `kubectl edit deployment -n kubecost`, which should open a text file containing your deployment in the namespace `kubecost`. Modify the CPU or RAM requests as desired, then save the file. In your terminal, your output will display CPU/RAM unit hours, total cost, and the cost difference (`diff` column).
--- a/using-kubecost/admission-controller.md
+++ b/using-kubecost/admission-controller.md
@@ -27,7 +27,7 @@ You may need to wait several minutes for the controller to activate. You can che
 
 ## Using the Admission Controller
 
-When a deployment is updated, the Kubernetes API will send requests containing deployment information to the Kubecost pod, which will then be read for number of replicas, CPU requests, and RAM requests to calculate a monthly estimate. That estimate will be reported back to the client making the update. If the Kubecost pod is unable to respond to this request, the deployment will be added to the cluster without any information being sent back.
+When a Deployment is updated, the Kubernetes API will send requests containing Deployment information to the Kubecost pod, which will then be read for number of replicas, CPU requests, and RAM requests to calculate a monthly cost estimate. That estimate will be reported back to the client making the update. If the Kubecost pod is unable to respond to this request, the deployment will be added to the cluster without any information being sent back.
 
 To validate, you can run `kubectl edit deployment -n kubecost`, which should open a text file containing your deployment. Modify the CPU or RAM requests as desired, then save the file. In your terminal, your output will display CPU/RAM unit hours, total cost, and the cost difference (`diff` column).
 
--- a/using-kubecost/admission-controller.md
+++ b/using-kubecost/admission-controller.md
@@ -1,0 +1,50 @@
+# Admission Controller
+
+{% hint style="info" %} The Admission Controller is a beta feature. Please read the documentation carefully. {% endhint %}
+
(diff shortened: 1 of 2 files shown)


bstuder99:
One thing I'd like to check in about is the location of this article in the left nav. Is Additional Configuration a better location for this? We don't really have a configurable Kubecost feature like this, so there isn't a lot of precedent for where to put this

srpomeroy:
Yes, `Additional Configuration` is a better place than the root.

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,50 @@
+# Admission Controller
+
+{% hint style="info" %} The Admission Controller is a beta feature. Please read the documentation carefully. {% endhint %}
+
+Kubecost's Admission Controller is a `kubectl` integration which leverages Kubecost's Predict API to display expected future costs for changes made to CPU and RAM in your workloads. This can be used to anticipate future spend before resources have been allocated.
```
michaelmdresser, line 5: > is a `kubectl` integration

This is not true. Kubecost's Admission Controller is a Kubernetes ["Admission Controller"](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/). The reason the demo you've probably seen uses `kubectl` is because `kubectl` is the most common/agnostic method of interacting with K8s. Admission Controllers affect all interactions with K8s, not just those which originate from `kubectl`.

The Kubecost Admission Controller currently lives within the cost-analyzer Pod today and, when enabled, hooks in to the K8s API server and instructs K8s to send all Deployment changes for review ("validation"). This behavior is captured well in the start of the following paragraph.

Non-kubectl interactions would still receive the cost prediction data and be able to display it if they wish -- it comes back as part of an "admission review". `kubectl` conveniently displays relevant parts of that review by default.

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Installing the Admission Controller
+
+### Prerequisites
+
+Before installing the Admission Controller, make sure you have [`kubectl`](https://kubernetes.io/docs/reference/kubectl/) installed.
+
+### Installation
+
+The following command will install the Admission Controller to your existing Kubecost application:
```
michaelmdresser, line 17:
```suggestion
The following command will enable the Admission Controller in your existing Kubecost application:
```

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,50 @@
...
+  --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
+  --namespace kubecost --create-namespace \
+  --set kubecostAdmissionController.enabled=true
+```
+
+You may need to wait several minutes for the controller to activate. You can check the status of the Admission Controller with `kubectl get service -n kubecost`, when the Admission Controller has been installed in the default `kubecost` namespace. Look for `webhook-server` to confirm a successfull install.
+
+## Using the Admission Controller
+
+When a deployment is updated, the Kubernetes API will send requests containing deployment information to the Kubecost pod, which will then be read for number of replicas, CPU requests, and RAM requests to calculate a monthly estimate. That estimate will be reported back to the client making the update. If the Kubecost pod is unable to respond to this request, the deployment will be added to the cluster without any information being sent back.
```
michaelmdresser, line 30:
```suggestion
When a Deployment is updated, the Kubernetes API will send requests containing Deployment information to the Kubecost pod, which will then be read for number of replicas, CPU requests, and RAM requests to calculate a monthly cost estimate. That estimate will be reported back to the client making the update. If the Kubecost pod is unable to respond to this request, the deployment will be added to the cluster without any information being sent back.
```

This is an excellent summary.

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,50 @@
...
+
+To validate, you can run `kubectl edit deployment -n kubecost`, which should open a text file containing your deployment. Modify the CPU or RAM requests as desired, then save the file. In your terminal, your output will display CPU/RAM unit hours, total cost, and the cost difference (`diff` column).
+
+This Admission Controller bundles an SSL cert/key pair with Kubecost that is publicly available. This allows the Admission Controller to operate over HTTPS, which is required by Kubernetes. For deploying to production, see the following section.
```
michaelmdresser, line 34: This bundling just changed. See https://github.com/kubecost/kubecost-cost-model/pull/1818 -- Thomas N is a good person to reach out to for clarifying installation.

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Deploying to production and namespaces other than `kubecost`
```
thomasvn, line 36: Concerning Michael's comment above about changes in bundling, yes he's right! The Admission Controller no longer bundles an SSL cert/key pair. You have to generate it yourself. My suggestion is:

Delete the following text:

```txt
This Admission Controller bundles an SSL cert/key pair with Kubecost that is publicly available. This allows the Admission Controller to operate over HTTPS, which is required by Kubernetes. For deploying to production, see the following section.
```

Remove the entire section `Deploy to production and namespaces other than kubecost`

Then move the following code example up into the `Installation` section:

```sh
git clone https://github.com/kubecost/cost-analyzer-helm-chart.git \
cd cost-analyzer-helm-chart/cost-analyzer/scripts \
./create-admission-controller-tls.sh <namespace-to-install-kubecost-in>
```

using-kubecost/admission-controller.md
bstuder99: Thanks, but this is loosely taken from our blog post, so I'm pretty sure you wrote this

using-kubecost/admission-controller.md
bstuder99: Thanks for clarifying this. Can you review the adjusted code block just to make sure I formatted it correctly? Also, while the user is able to install the Admission Controller in a namespace different from `kubecost`, does it still need to be namespace Kubecost is installed in? ie, in the install command, should it be:
```
helm upgrade --install kubecost \
  --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
  --namespace <namespace-to-install-kubecost-in> --create-namespace \
  --set kubecostAdmissionController.enabled=true \
git clone https://github.com/kubecost/cost-analyzer-helm-chart.git \
cd cost-analyzer-helm-chart/cost-analyzer/scripts \
./create-admission-controller-tls.sh <namespace-to-install-kubecost-in>
```
Where both namespace values need to match? Our default install command creates the namespace kubecost, so should the Admission Controller namespace be kubecost as well?

using-kubecost/admission-controller.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+helm upgrade --install kubecost \
+  --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
+  --namespace kubecost --create-namespace \
+  --set kubecostAdmissionController.enabled=true \
+git clone https://github.com/kubecost/cost-analyzer-helm-chart.git \
+cd cost-analyzer-helm-chart/cost-analyzer/scripts \
+./create-admission-controller-tls.sh <namespace-to-install-kubecost-in>
+```
```
thomasvn, line 29: Although I don't think the admission controller needs to be in the same namespace as Kubecost, let's just make it that way in this doc for simplicity. Is that ok?

Here's my suggestion:

```
# Setup TLS key/certificate for the Admission controller
$ git clone https://github.com/kubecost/cost-analyzer-helm-chart.git
$ cd cost-analyzer-helm-chart/cost-analyzer
$ kubectl create namespace kubecost
$ ./scripts/create-admission-controller-tls.sh kubecost

# Install Kubecost with AdmissionController enabled
$ helm upgrade --install kubecost \
  --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
  --namespace kubecost --create-namespace \
  --set kubecostAdmissionController.enabled=true \
  -f values.yaml
```

chipzoller: Asking users to clone an entire git repo for purposes of executing a single bash script is way overkill. This can be accomplished much more simply:

```sh
curl -s https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/develop/cost-analyzer/scripts/create-admission-controller-tls.sh | bash
```

Note that the script defaults to a param of `kubecost` if one is not supplied, therefore no arguments are required if the default Namespace is selected. This script is predicated on the Namespace preexisting, so if used as the first step then the user will need to be instructed on how to create a new Namespace.

using-kubecost/admission-controller.md
chipzoller: +1 to the above from Michael. Just adding that the returned messages displayed by the API server do not even depend on using the `kubectl` program; they are returned in the payload regardless.

chipzoller:
If you would link this PR to the issue it is intended to close, the issue will close automatically when the PR is closed. See GitHub's [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for details. I have gone ahead and made this change for you to illustrate how it's done.

using-kubecost/admission-controller.md
chipzoller: I would recommend not asking users to hand-edit the Kubecost Deployment but rather some other test Deployment in their environment. A simple tutorial here would be in order to show users the effects of a Deployment creation and subsequent update.

chipzoller:
Xref PR to improve script UX: https://github.com/kubecost/cost-analyzer-helm-chart/pull/2646

using-kubecost/admission-controller.md
chipzoller: Also, if using this approach the `-f values.yaml` flag will need to be omitted as this assumes the values file is available locally which, if the repo hasn't been cloned locally, it won't be. Since the `--set` directive is used, assuming no other value overrides are provided this isn't needed.

thomasvn: @chipzoller Agree with all of the above! Currently the `create-admission-controller-tls.sh` attempts to find and replace the string `${CA_BUNDLE}` in the `values.yaml` file. Hence why these commands clone the repo and include `-f values.yaml`

If we change the way the script works, we can in turn make this documentation simpler? Thoughts on best practices here Chip?

chipzoller: I clearly didn't pay attention, sorry about that. I'll make a slight improvement to the script so it'll optionally just print out the encoded CA to use, but otherwise just disregard my comments here.

michaelmdresser (approved):

PR #761, opened 2023-10-12
More docs renames

bstuder99:
## Proposed Changes

Renames a couple docs to be more descriptive


api-allocation.md
```diff
@@ -22,7 +22,7 @@ If `true`, sum the entire range of sets into a single set. Default value is `fal
 {% endswagger-parameter %}
 
 {% swagger-parameter in="path" name="idle" type="boolean" required="false" %}
-If `true`, include idle cost (i.e. the cost of the un-allocated assets) as its own allocation. (See [special types of allocation](https://docs.kubecost.com/apis/apis-overview/allocation#special-types-of-allocation).) Default is `true.`
+If `true`, include idle cost (i.e. the cost of the un-allocated assets) as its own allocation. (See [special types of allocation](api-allocation.md).) Default is `true.`
```
srpomeroy, line 25: This is linked to itself. It was previously a link to an anchor within this doc.

Should it be the following?
```suggestion
If `true`, include idle cost (i.e. the cost of the un-allocated assets) as its own allocation. (See [special types of allocation](#special-types-of-allocation).) Default is `true.`
```

srpomeroy (changes requested):

api-allocation.md
bstuder99: Sorry, I'm renaming the allocation.md file to be more descriptive (api-allocation.md to stop confusing it with our Allocation UI doc), so it will look like api-allocation.md#special-types-of-allocation, correct?

srpomeroy: I think the anchor should suffice as long as it's in the same doc.

srpomeroy (approved):

PR #767, opened 2023-10-17
Creates document on cost-analyzer ClusterRoles

bstuder99:

### DIFF
--- a/install-and-configure/advanced-configuration/user-management/kubecost-cluster-roles.md
+++ b/install-and-configure/advanced-configuration/user-management/kubecost-cluster-roles.md
@@ -1,0 +1,49 @@
+# Kubecost Cluster Roles
+
+Kubecost requires read only RBAC permissions on most cluster resources in order to build a granular cost-model for all resources. `Role` can be set to make changes in your namespace, while `ClusterRole` is required to make changes across the cluster (and therefore in all namespaces).
+
+`cluster-admin` is required to install Kubecost. However, this role is not required to modify the deployment afterwards.
+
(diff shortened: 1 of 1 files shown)


bstuder99:
This is just the first pass on RBAC cluster roles, as our current documentation doesn't touch this topic too heavily. I'd like to expand this based on user requests over time or internally where we feel content is lacking. Our [SSO/SAML/RBAC doc](https://docs.kubecost.com/install-and-configure/advanced-configuration/user-management) mentions these three predefined RBAC roles: `admin`, `readonly`, and `editor`.

bstuder99:
1. The idea was to be able to transcribe the content from the cluster-roles-template.yaml into a more readable, or at least condensed, format. Content may be subject to change like the rest of our documentation, as long as PRs affecting the original .yaml are noted during release, we should be able to make consistent updates to this doc.
2. Noted
3. Maybe you should open that issue with the specifics as to what is outdated/incorrect? Would like to know myself so I can update this doc.

chipzoller:
> 1. The idea was to be able to transcribe the content from the cluster-roles-template.yaml into a more readable, or at least condensed, format. Content may be subject to change like the rest of our documentation, but as long as PRs affecting the original .yaml are noted during release, we should be able to make consistent updates to this doc.

I understand the idea but I think there's a more sustainable, less fragile way to do this which doesn't involve near-immediate obsolescence. For example:

"Kubecost requires `get`, `list`, and `watch` permissions over many common Kubernetes Pod and Pod controller resources such as Pods, Deployments, StatefulSets as well as other resources which factor into to cost analysis such as Namespaces, Nodes, Events, etc."

The "etc" part I'd list out all the specific resources. This provides the necessary information without showing the exact implementation details which are subject to change. End with providing a link to the actual ClusterRole(s) and state something generic like, "the current set of permissions can be found at the following location."

chipzoller:
Created https://github.com/kubecost/cost-analyzer-helm-chart/issues/2658

jessegoodier (approved):
looks good to me.

PR #778, opened 2023-10-20
Adds cluster Roles to Summary.md

bstuder99:


jessegoodier (approved):

thomasvn:
@bstuder99 I recommend moving this doc to the "Architecture" section. All our existing RBAC docs focus on app-level RBAC (e.g. Okta groups/roles, and how we interface with those claims). However, this doc focuses on k8s-level RBAC (e.g. [ClusterRole](https://kubernetes.io/docs/reference/access-authn-authz/rbac/))

Therefore, I think these are two separate topics.

PR #779, opened 2023-10-23
Relocates docs files into Using Kubecost UI folders

bstuder99:
## Proposed Changes

This PR focuses on relocating files into the Using Kubecost folders, which comprises mainly of UI-related docs.

@thomasvn Also relocates Cluster Roles article into architecture folder


bstuder99:
Going to fix links first

PR #783, opened 2023-10-24
Updates Budget API doc

bstuder99:
## Proposed Changes

Updates the existing Budget API doc with 1.107 endpoints, while moving the existing content into a "new" deprecated doc.


apis/apis-overview/budget-api.md
```diff
@@ -1,8 +1,8 @@
-# Budget API
+# Budget API (v2)
 
-The Budget API allows you to create, update, and delete recurring budget rules to control your Kubernetes spending. Weekly and monthly budgets can be established to set limits on cost spend.
+The Budget API allows you to create, update, and delete recurring budget rules to control your Kubernetes spending. Weekly and monthly budgets can be established to set limits on cost spend, with the option to configure alerts for reaching specified budget thresholds via email, Slack, or Microsoft Teams.
```
avrodrigues5, line 3: Can we name it actions instead of alerts as that the word used.

apis/apis-overview/budget-api.md
```diff
@@ -23,37 +23,60 @@ The interval that the budget will reset with (either `weekly` or `monthly`).
...
+                "end": ""
+            },
+            "currentSpend":
         }
 }
 ```
 {% endswagger-response %}
 {% endswagger %}
 
-{% swagger method="get" path="/model/budget/recurring/list" baseUrl="http://<your-kubecost-address>" summary="Get recurring budget rule(s)" %}
```
avrodrigues5, line 56: there's no concept of recurring anymore, you could drop it to just "budget" -> so it can be Get all  Budgets

apis/apis-overview/budget-api.md
```diff
@@ -23,37 +23,60 @@ The interval that the budget will reset with (either `weekly` or `monthly`).
...
+                    "<name of namespace of cluster>"
+                ]
+            },
+            "kind": "",
+            "interval": "",
+            "intervalDay": ,
+            "spendLimit": ,
+            "actions": [
+                {
+                    "amount": 0,
```
avrodrigues5, line 56: Can you remove Amount here ? you cant give both amount and percentage it will error out!

avrodrigues5 (approved):
other than minor corrections LGTM on budget v2 side.

apis/apis-overview/budget-api.md
bstuder99: I know Alerts is an existing feature, but saying the user can configure "an action for email" doesn't exactly convey what is happening here. We elaborate later in the article how this is done through the actions param

apis/apis-overview/budget-api.md
bstuder99: Good catch, forgot to update this endpoint

apis/apis-overview/budget-api.md
bstuder99: Yep, good call out

PR #785, opened 2023-10-24
OIDC and Readonly helm configs

thomasvn:
## Related issue #

Documents https://github.com/kubecost/kubecost-cost-model/pull/1809

## Proposed Changes

Adds info about how `.Values.oidc.enabled=true` and `.Values.readonly=true` are compatible.


user-management-oidc.md
```diff
@@ -96,6 +96,16 @@ The JWT ID token must contain a field (claim) named `hd` with the desired domain
 
 To remove a previously set Helm value, you will need to set the value to an empty string: `.Values.oidc.hostedDomain = ""`. To validate that the config has been removed, you can check the `/var/configs/oidc/oidc.json` inside the cost-model container.
 
+## Readonly mode
```
chipzoller, line 99:
```suggestion
## Read-only mode
```

user-management-oidc.md
```diff
@@ -96,6 +96,16 @@ The JWT ID token must contain a field (claim) named `hd` with the desired domain
...
+
+Kubecost's OIDC supports readonly mode. This leverages OIDC for authentication, then assigns all authenticated users as readonly users.
```
chipzoller, line 101:
```suggestion
Kubecost's OIDC supports read-only mode. This leverages OIDC for authentication, then assigns all authenticated users as read-only users.
```

chipzoller (commented):
Typically stylized as "read-only" regardless of what the code-specific setting may be.

bstuder99:
After Chip's suggestions, LGTM

user-management-oidc.md
thomasvn: Thanks!

chipzoller (dismissed):

chipzoller:
Broken link

bstuder99:
@chipzoller I have that link fixed in the PR #787, should we merge that first? Seems like a leftover from yesterday's repo cleanup

chipzoller:
Yes, it's best to merge that first, update the branch to let CI checks run again, and then merge if clear.

bstuder99:
Is there a way to force the link checker to run again? I believe it only updates when commits are added to the individual PR it's running on. So my changes from #787 won't cause it to run again even though it should resolve the error. Could be wrong on this, but that's the behavior I've observed.

srpomeroy:
> Is there a way to force the link checker to run again? I believe it only updates when commits are added to the individual PR it's running on. So my changes from #787 won't cause it to run again even though it should resolve the error. Could be wrong on this, but that's the behavior I've observed.

You should be able to click on the details like on the failed check and then kick off a re-run.

srpomeroy:
> ### Errors in install-and-configure/install/etl-backup/query-service-replicas.md
>
>* [404] [https://docs.kubecost.com/install-and-configure/install/multi-cluster/thanos-setup/long-term-storage#step-1-create-object-store.yaml](https://docs.kubecost.com/install-and-configure/install/multi-cluster/thanos-setup/long-term-storage#step-1-create-object-store.yaml) | Failed: Network error: Not Found

bstuder99:
Ok, I've manually fixed the link and the conflict, we should be good here

bstuder99 (approved):

bstuder99 (approved):

PR #788, opened 2023-10-25
Update auto-request-sizing.md

kwombach12:
fix small typo - assuming we meant AKS here and not AWS


bstuder99:
In this context, we aren't referring to Azure Kubernetes Service, it's specifically [Kops by AWS](https://aws.amazon.com/blogs/compute/kubernetes-clusters-aws-kops/). Why we support this and not Azure I'm not too sure, but this is what's been explained to me.

bstuder99:
This is the same functionality in our [Cluster Controller](https://docs.kubecost.com/install-and-configure/advanced-configuration/cluster-controller) doc.

kwombach12:
@bstuder99 ah ok that makes sense. I will nuke this PR

PR #803, opened 2023-11-03
Fixes image not appearing, misc. improvements

bstuder99:


thomasvn:
@bstuder99 Thanks for the quick fix! I don't see that we have a Gitbook Preview link available for this PR. But let's merge and fix again if needed.

thomasvn (approved):

PR #807, opened 2023-11-07
Additional accessibility updates, docs cleanup

bstuder99:


install-and-configure/install/helm-install-params.md
```diff
@@ -1,39 +1,42 @@
...
+Then run your install command:
+
+```
+$ helm install kubecost cost-analyzer \
+    --repo https://kubecost.github.io/cost-analyzer/ \
+    --namespace kubecost --create-namespace \
+    --values values.yaml
+```
+
+## Method 3: Use [_values.yaml_](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml) from the Kubecost Helm chart repository.
```
srpomeroy, line 38: Do we need to keep this? The referenced values file contains the default values applied when running a helm install. You don't need to refer to that file specifically unless the user has modified it.

srpomeroy (changes requested):

install-and-configure/install/helm-install-params.md
srpomeroy: @jessegoodier @chipzoller What are your thoughts?

install-and-configure/install/helm-install-params.md
```diff
@@ -10,9 +10,9 @@ For example, you can only pass a product key if that is all you need to configur
 $ helm install kubecost cost-analyzer \
     --repo https://kubecost.github.io/cost-analyzer/ \
     --namespace kubecost --create-namespace \
-    --set kubecostProductConfigs.productKey.key="123"
+    --set kubecostProductConfigs.productKey.key="123" \
+    --set kubecostProductConfigs.productKey.enabled=true
     ...
-```
```
srpomeroy: Looks like the suggestion removed the closing back ticks.

bstuder99: Fixed

PR #831, opened 2023-11-30
update airgapped images

jessegoodier:
updating airgapped images doc to reflect current images.

@bstuder99 for your normal expertise on making this look good.

@chipzoller for any concerns


jessegoodier:
> Added my edits into the doc, thanks for cleaning this up

great edits- thanks!

PR #835, opened 2023-12-05
Creates Cloud Cost Trends API doc, updates Allocation Trends content

bstuder99:
## Proposed Changes

Dev PR at https://github.com/kubecost/kubecost-cost-model/pull/1937


bstuder99:
Doc won't render as intended in GH's MD preview, but if you select Show all checks, GitBook has a preview that will show you what the doc will display as

apis/apis-overview/cloud-cost-api/cloud-cost-trends-api.md
```diff
@@ -0,0 +1,92 @@
...
+            "start": "2023-11-22T00:00:00Z",
+            "end": "2023-11-29T00:00:00Z"
+        }
+```
+
+Trend values are converted into percentages in the Kubecost Cloud Costs Explorer page, calculated based on your current query. Trends will be presented in the rightmost column, next to your Total cost. The `window` parameter is determined by your selected date range in the top right of the page. The default is Last 7 days (`window=7d`). The equation `value*100` is used to provide percentages.
+
+## Using the `names` parameter
+
+`names` is a mandatory parameter which determines the sequence of items returned, based on whatever the query is aggregating by. For example, when `aggregateBy=provider`, the user should provide a comma-separated list of all items they which to see trend values for in this category. In this case, they should provide `names=AWS,GCP,Azure` to receive a list of trend values for all three providers. `service` is the default value for `AggregateBy`, so if the user does not provide a value for `aggregateBy`, they must still use the names parameter to list all services requested. For example, `names=AmazonEC2,AmazonS3,Microsoft.Compute...`.
```
nickcurie, line 92: The actual parameter for aggregation is `aggregate` not `aggregateBy`

apis/apis-overview/cloud-cost-api/cloud-cost-trends-api.md
```diff
@@ -0,0 +1,92 @@
...
+{% endswagger-response %}
+{% endswagger %}
+
+## Calculating trend value
+
+The Trends API determines changes in resource cost usage over time based on the interval set `window` parameter and provides that information via the schema field `value`. Cost usage for the current `window` sampled will be compared with `comparisonWindow`, the window directly before the current window of the same size interval. For example, for `window=3d`, Kubecost will output cost usage for the past three days compared to cost usage of the three days before the start of the window. This means a total of six days of cloud cost data are sampled in order to provide an accurate value.
+
+The equation for calculating `value` is: `value=window/comparisonWindow - 1`
+
+Receiving a positive `value` means your `window` has increased compared to `comparisonWindow`. A negative `value` means spending has decreased.
```
nickcurie, line 62: Should this be something like "Receiving a positive `value` means your spending has increased in `window` compared to `comparisonWindow` ..." for clarity? As it is reads kind of strange to me

nickcurie (changes requested):
This looks good, just one or two quick things

nickcurie (approved):

apis/apis-overview/cloud-cost-api/cloud-cost-trends-api.md
```diff
@@ -0,0 +1,91 @@
...
+{% swagger-parameter in="path" name="window" required="true" type="string" %}
+Duration of time over which to query. Compares cost usage of window to cost usage window of equal size directly preceding it. Accepts all standard Kubecost window formats (See our doc on using [the `window` parameter](https://docs.kubecost.com/apis/apis-overview/assets-api#using-window-parameter)).
+{% endswagger-parameter %}
+
+{% swagger-parameter in="path" name="names" required="true" type="string" %}
+Determines order sequence of queried items via comma-separated list. Dependent on the value of `aggregateBy` to list items. See more [below](cloud-cost-trends-api.md#using-the-names-parameter).
+{% endswagger-parameter %}
+
+{% swagger-parameter in="path" name="aggregate" type="string" required="false" %}
+Field by which to aggregate the results. Accepts: `invoiceEntityID`, `accountID`, `provider`, `service`, and `label:<name>`. Supports multi-aggregation using comma-separated lists, such as `aggregate=accountID,service`. Defaults to `service`.
```
nickcurie, line 16: I'm realizing now that we actually support more than this and that the default is not actually `service` but an empty string. Using an empty string for aggregate would cause us to aggregate by item. 

As for a list of acceptable aggregate parameters, we can take `invoiceEntityID`, `accountID`, `provider`, `providerID`, `category`, `service`, and `label:<name>`

nickcurie (changes requested):

apis/apis-overview/cloud-cost-api/cloud-cost-trends-api.md
bstuder99: So if they don't provide a value for aggregate, they will have to list every single line item in Names for a complete breakdown of their trends?

nickcurie: Yes they would

PR #836, opened 2023-12-05
Account mapping in Kubecost UI

bstuder99:
## Proposed Changes

Docs for:
* https://github.com/kubecost/kubecost-cost-model/pull/1944
* https://github.com/kubecost/cost-analyzer-helm-chart/pull/2796


install-and-configure/install/cloud-integration/README.md
```diff
@@ -71,6 +71,20 @@ The Kubecost Helm chart provides values that can enable or disable each cloud pr
...
 
+## Generating readable cloud account names
+
+Often an integrated cloud account name may be a series of random letter and numbers which do not reflect the account's owner, team, or function. Kubecost allows you to rename cloud accounts to create more readable cloud metrics in your Kubecost UI. After you have successfully integrated your cloud account (see above), you need to manually edit your *values.yaml* and provide the original account name and your intended rename:
```
Sean-Holcomb, line 76: Can we refer to the un readable string as "Account ID" and the readable one as "Account Name". This distinction is present in the Focus spec I think worth while to preserve.

Sean-Holcomb (dismissed):
Looks good, with one note

PR #838, opened 2023-12-07
Renames image files to be more descriptive

bstuder99:
## Proposed Changes

Likely not an exhaustive fix, but does several things:
* Renames all iterations of `image...png` found in the repo to be more descriptive
* Deletes duplicate image files or image files found not to exist anywhere in documentation

I manually checked each updated image so links should all still be functional (even if the Broken Link Checker doesn't scan for these). I still need to comb through and find more unused image files, as some are replaced with updated images and are never deleted.


bstuder99:
Looks like there's some funky business with the link checker, will resolve those ASAP

srpomeroy (approved):

PR #854, opened 2024-01-04
2.0 changes: Federated ETL and Aggregator docs

michaelmdresser:
## Related issue #

- See release notes: https://github.com/kubecost/cost-analyzer-helm-chart/pull/2898


michaelmdresser:
Hey @jessegoodier I'm still working on this, but how do you think we should be updating Thanos documentation? I assume we're going to have to document the ETL Utils setup here.

install-and-configure/install/etl-backup/query-service-replicas.md
bstuder99, line 1: In a separate PR I'm drafting, I considered removing this doc altogether. If this is going to be supported to a certain point for users on older versions of Kubecost, it might make sense to keep the content of this doc on top of the disclaimer you added.

michaelmdresser: This is a good point. Feel free to close this PR in favor of yours, it's better to centralize work.

I think more broadly we need to decide how to handle documentation for features that remain in-support for pre-2.0 version of KC but out-of-support post-2.0. Do you have any ideas?

bstuder99:
Is there a reason the High Availability doc is being removed also? I would consider adding a similar notice at the top of the page as the QSR doc, at least until we decide to no longer officially support it or give a sunset phase for users.

bstuder99:
Just a heads up, we have a number of 404s related to page content and Helm charts being moved around.

michaelmdresser:
> Is there a reason the High Availability doc is being removed also? I would consider adding a similar notice at the top of the page as the QSR doc, at least until we decide to no longer officially support it or give a sunset phase for users.

The PR was super off-the-cuff; I think you're right that a notice rather than a full deletion is a better idea.

michaelmdresser:
@bstuder99 my apologies, but I've let this PR rot for a long time and don't have time to pick it up again before the 2.0 release. I'm not sure where we ended up on delete old docs vs. leave with big deprecated banner (or something), so I'm pretty sure this PR should be closed. I'll let you make the call, there might be some ideas here that should be ported over to the current state of the 2.0-changes branch.

bstuder99:
I have a couple things I'll have to fix up, ie TODO comments left in the docs, but there's valuable content in here I don't want to scrap. Can you verify the content you wrote for the Aggregator doc is all accurate?

michaelmdresser:
@bstuder99 If we ignore the changes in this PR for QSR and HA, the rest of my writing (and modifications) I'm pretty happy with here

bstuder99:
Ok solid. The QSR/HA docs will likely be deleted after the next major update (as I'm pretty sure we only support the two most recent product releases at any time).

PR #859, opened 2024-01-12
renamed quick-start doc to more accurately reflect that it's for envs…

mmurph3:
… using IRSA/OIDC


bstuder99:
@jessegoodier for some visibility

bstuder99:
Two broken links because the file name is being changed which I can correct, just want to finalize our intended article title before I do that

PR #872, opened 2024-01-29
Further details as to why Aggregator StatefulSet must be manually deleted in some upgrades

thomasvn:
## Related issue

Documents https://github.com/kubecost/cost-analyzer-helm-chart/pull/2964

## Proposed Changes

Provides a detailed explanation for why the StatefulSet must be deleted. And provides and example command that can be run to resolve the situation.


thomasvn:
Closing this PR, I mean to create a PR which merges into `2.0-changes`

PR #875, opened 2024-01-31
Update getting-started.md

kwombach12:
update to remove reference to SSO being enterprise only

### DIFF
--- a/install-and-configure/install/getting-started.md
+++ b/install-and-configure/install/getting-started.md
@@ -39,7 +39,7 @@ Establishing an ingress controller will allow for control of access for your wor
 ### SSO/SAML/RBAC/OIDC
 
 {% hint style="info" %}
-SSO/SAML/RBAC/OIDC are only officially supported on Kubecost Enterprise plans.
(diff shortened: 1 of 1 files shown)


install-and-configure/install/getting-started.md
```diff
@@ -39,7 +39,7 @@ Establishing an ingress controller will allow for control of access for your wor
 ### SSO/SAML/RBAC/OIDC
 
 {% hint style="info" %}
-SSO/SAML/RBAC/OIDC are only officially supported on Kubecost Enterprise plans.
```
saweber, line 42: This change is an improvement.

nit: Maybe just say SAML/OIDC? RBAC is really a feature layered on top of a SAML or OIDC implementation.

install-and-configure/install/getting-started.md
```diff
@@ -39,7 +39,7 @@ Establishing an ingress controller will allow for control of access for your wor
 ### SSO/SAML/RBAC/OIDC
```
saweber, line 39: should this header name change as well?

bstuder99:
@kwombach12 I have a separate PR #877, can we close this and push that one instead?

PR #877, opened 2024-01-31
Fixes banners in SSO/SAML/RBAC content

bstuder99:


thomasvn:
I believe we may want to adjust the wording because SAML/OIDC are considered enterprise features as well (as far as I understand).

What would be considered "free" is setting up your own SSO (i.e. [oauth proxy](https://github.com/oauth2-proxy/oauth2-proxy)) which sits in front of the Kubecost app. cc @kwombach12

jessegoodier:
Let me verify. thanks for this @bstuder99

PR #881, opened 2024-02-02
Delete storage.md as it's outdated and now implicitly required.

thomasvn:
## Related issue

N/A


bstuder99:
Were you intending to delete storage.md itself or just this reference to it?

thomasvn:
I was going to delete the file itself, is that ok? Or do you prefer I just remove all references to it.

bstuder99:
I'm not against that, as I think we'll still preserve it in a 1.x branch, but it would still need to be delisted from summary.md. I'll make the edit.

bstuder99:
Oh I see you already made the edit. Let me review the link checker before we merge

bstuder99 (approved):

PR #886, opened 2024-02-06
Removes Cloud Cost configuration section

bstuder99:
## Proposed Changes

Simplifies Cloud Cost Explorer doc, as there is no longer any need to configure/install CloudCost flags ahead of time. This is now the default, any users with older Kubecost versions can consult the 1.x docs.


bstuder99:
Broken link goes to some Amazon EKS content, which is fixed in another PR which still needs to be merged

using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer.md
```diff
@@ -8,46 +8,6 @@ The Cloud Cost Explorer is a dashboard which provides visualization and filterin
...
-       # format labels as comma separated string (ex. "label1,label2,label3")
-       labels: ""
-     topNItems: 1000
-```
-
-Enabling Cloud Cost is required. Optional parameters include:
-
-* `labelList.labels`: Comma-separated list of labels; empty string indicates that the list is disabled
-* `labelList.IsIncludeList`: If true, label list is a white list; if false, it is a black list
-* `topNItems`: number of sampled "top items" to collect per day
```
thomasvn, line 34: This doesn't have to be addressed in this PR, but for a future PR.

I agree that we should delete this section since these are now on by default. However it would be good to still find a place to document `labels`, `IsIncludeList`, and `topNItems` since those are still configurable via the Helm chart. https://github.com/kubecost/cost-analyzer-helm-chart/blob/7cb64ae627778d9eac917ad039a0507bbc81a375/cost-analyzer/templates/_helpers.tpl#L1064-L1077

thomasvn (approved):

using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer.md
bstuder99: There is a thread on Slack about this, apparently these actually don't do anything as of v2.0

PR #894, opened 2024-02-09
Makes Aggregator API endpoints only documented endpoints in 2.0 docs

bstuder99:
## Proposed Changes

Some time before v2.0 we updated some API docs to include Aggregator-specific endpoints and content as an alternative. Now that Aggregator is the default, these will be the default endpoints and we are removing 1.x API endpoints.


thomasvn:
Looks good. Should we leave the 1.x API endpoints in for a few more releases? It may still be useful to some.

bstuder99:
Those will be saved in the 1.x version, plan is to launch that soon so we can at least hold off on merging this PR until it's up.

PR #896, opened 2024-02-09
document service accounts

saweber:
## Related issue #
https://docs.google.com/document/d/14C3byaE-hFTnDFoIMcONIqwE9GrAQjjvGE_91IZ4cpM/edit?usp=sharing
https://kubecost.atlassian.net/browse/SELFHOST-1173

## Proposed Changes
Adds documentation for service accounts. The 'how to create an application key' section will need to be revisited once the FE work is completed.


bstuder99:
I've listed the article in Summary.md and updated the frontend section

srpomeroy:
According to the PRD, only a user with admin rights can create a service account. Can we add this detail to the documentation as well?

bstuder99:
Sean, does that admin right need to be configured through the Teams feature? Or are there other ways to configure it? Sorry, don't have the PRD in front of me

srpomeroy:
> Sean, does that admin right need to be configured through the Teams feature? Or are there other ways to configure it? Sorry, don't have the PRD in front of me

PRD is linked in the opening PR comment.

The admin role would either be assigned via group membership as defined in the OIDC/SAML helm config, or via the new Teams config.

saweber:
The user is required to be an admin, but it does not have to be from teams.

If SAML RBAC or OIDC RBAC are enabled:
If the user is configured as an admin through helm based RBAC, that will work as well.

If SAML or OIDC are enabled without RBAC, any authorized user will be able to create service accounts

If SAML and OIDC are disabled, any user will be able to create service accounts (though these service accounts won't be used, because there is no authentication on the kubecost installation at all).

bstuder99:
Added a final commit which adds more details to Steven's commit

PR #909, opened 2024-02-16
Describing `DB_BUCKET_REFRESH_INTERVAL`, and when it needs to be tuned.

thomasvn:
## Proposed Changes

First attempt at describing this env var, and in what scenarios it needs to be tuned. @ameijer @michaelmdresser Would love feedback on whether it's an accurate description or not.

### DIFF
--- a/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
+++ b/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
@@ -156,7 +156,7 @@ When deploying the Aggregator as a StatefulSet, it is possible to perform a rese
 
 ### Aggregator not displaying any data to frontend after several hours
 
-One reason you may not see data in the frontend yet is because the Aggregator is still in the process of reading all data from the federated store bucket, and processing the data into SQL tables.
+One reason you may not see data in the frontend yet is because the Aggregator is processing all your ETL files in the federated store bucket into SQL tables.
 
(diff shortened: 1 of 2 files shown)


jessegoodier (approved):
Thank you!

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -138,3 +142,33 @@ As `ETL_DAILY_STORE_DURATION_DAYS` increases, the amount of time it will take
...
+
+One reason you may not see data in the frontend yet is because the Aggregator is processing all your ETL files in the federated store bucket into SQL tables.
+
+If you are seeing a lot of the following logs, it could be an indicator that your `.Values.kubecostAggregator.env.DB_BUCKET_REFRESH_INTERVAL` may be set too low, causing the Aggregator to continuously restart its data ingestion process:
+
+```txt
+INF asset worker context cancelled: context canceled
+INF allocation worker context cancelled: context canceled
+```
+
```
michaelmdresser, line 167: Maybe add something like this:

The default value of `DB_BUCKET_REFRESH_INTERVAL` is tuned for the first-install experience of Kubecost, meaning it prioritizes small data becoming available more quickly. As your data scale and retention increase and particularly once multiple clusters are set up, the value should be increased.

michaelmdresser (dismissed):

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: Oh that's good context! I've added that as part of the description of the env var.

bstuder99 (approved):

jessegoodier (approved):

PR #913, opened 2024-02-22
Adds section to Reports page troubleshooting for reports not properly displaying

bstuder99:


using-kubecost/navigating-the-kubecost-ui/saved-reports/reports.md
```diff
@@ -183,6 +185,79 @@ data:
...
 {% endcode %}
 
-3\. Ensure that the JSON string is successfully mapped to the appropriate configs
+3. Ensure that the JSON string is successfully mapped to the appropriate configs
+
+Navigate to the Reports page and ensure that the configured report parameters have been set by selecting the Report name.
+
+### Reports configured in the UI do not display properly after upgrading to v2.0+
+
+If your reports do not display properly after upgrading to v2.0+ from v1.x, you will need to manually copy your reports from a JSON file to from your KCM's `persistent-configs` PVC and move them into your `peristent-configs-aggregator-aggregator-0` PVCs.
```
michaelmdresser, line 194:
```suggestion
If your reports do not display properly after upgrading to v2.0+ from v1.x, you will need to manually copy your reports from a JSON file to from your KCM's `persistent-configs` PVC and move them into your `peristent-configs-aggregator-aggregator-0` PVC.
```

using-kubecost/navigating-the-kubecost-ui/saved-reports/reports.md
```diff
@@ -183,6 +185,79 @@ data:
...
+
```
michaelmdresser, line 195: I think it would be valuable to add a note like "this is an example for copying over just the Asset reports"

michaelmdresser (commented):
@teevans do you want this initial docs PR to go in and then we follow up with the scripted approach outlined in https://kubecost.atlassian.net/browse/SELFHOST-1227? Or do you want our docs to wait on the scripted approach?

using-kubecost/navigating-the-kubecost-ui/saved-reports/reports.md
bstuder99: I have this line later in the tutorial: "Compare both instances of `asset-reports.json`, which has a reduced file size post-upgrade, meaning it is not displaying properly in Kubecost's UI. Follow along this process of copying this example file with any affected reports." Does this cover it?

michaelmdresser (commented):
The doc is solid, but I realize it may be slightly obsolete because @williamkubecost put together https://github.com/kubecost/kubecost-utilities/tree/main/copy-reports to make this process easy for users. I'm sorry to ask: would you rewrite this doc to recommend that script?

PR #924, opened 2024-03-01
azure cloud integration using azure workload identity

mmurph3:


Sean-Holcomb (dismissed):
In general this looks good. Would it make sense to combine the two sections that update helm values, there isn't really much point in applying those values after step 3 if the secret is not in place. 

My other question is if we should include any instruction about the federated identity on how to create it and what permissions need to be applied to it.

mclean0328 (approved):
LGTM

PR #927, opened 2024-03-07
Removes the "Cloud Cost Diagnostic APIs" doc.

thomasvn:
## Proposed Changes

- Deletes the "Cloud Cost Diagnostic APIs" doc as much of the information is legacy.
- All CloudUsage APIs are deprecated, as this is no longer available in KC 2.0
- Reconciliation APIs are also deprecated, as reconciliation is now handled by the aggregator.
- `/model/etl/status` is an API that's still used, but can be moved to a different doc in the future. It's not often used by the user.

@Sean-Holcomb for confirmation?


bstuder99 (approved):
Thank you for going through and correcting this, I did not realize these changes were part of 2.0

PR #958, opened 2024-03-27
Misc. docs fixes for V2.2

bstuder99:
## Proposed Changes
 
* Updates that ContainerStats Pipeline is enabled by default
* Removes banners regarding multi-cluster Prom configs


jessegoodier (dismissed):
nice work

PR #959, opened 2024-03-27
Creates External Costs page

bstuder99:


using-kubecost/navigating-the-kubecost-ui/external-costs.md
```diff
@@ -0,0 +1,90 @@
...
+          type: RuntimeDefault
+        capabilities:
+          drop:
+          - ALL
+        readOnlyRootFilesystem: true
+        runAsNonRoot: true
+        runAsUser: 1000
+    folder: /opt/opencost/plugin
+    # leave this commented to always download most recent version of plugins
+    # version: <INSERT_SPECIFIC_PLUGINS_VERSION>
```
bstuder99, line 46: Is there a default version we can use, or does the user still need to visit opencost-plugin and find the most recent version?

nik-kc: There's no default, and there's no plan for doing any kind of LTS for any plugin versions.

Generally, users should keep this commented out though. They'll almost always need the latest version.

nik-kc:
It may also be useful to mention that external costs are mapped to fields in the [spec.pdf](https://github.com/kubecost/docs/files/14780699/spec.pdf).

bstuder99:
> It may also be useful to mention that external costs are mapped to fields in the [spec.pdf](https://github.com/kubecost/docs/files/14780699/spec.pdf).

Can I get some more context to what this file is? Is there somewhere in the GH org where this is hosted? Don't want to just give the user a hyperlink to a downloadable.

PR #978, opened 2024-04-07
Add use case on proactive cost controls

chipzoller:
Signed-off-by: chipzoller <chipzoller@gmail.com>

## Proposed Changes

Adds a new page under the Using Kubecost section called "Proactive Cost Controls" which describes a use case on how to prevent cost overruns before they occur. This links to a Kubecost blog post. Further links on the predictions API page have been added and a new section to the budgets page to remain consistent with page styling.


bstuder99:
Thank you for the content updates. To be honest, I think it needs some expansion to exist as its own technical doc; at the moment, I think the use cases described in here should be applied directly into the Budgets/Prediction API docs. What I think we could also use is a step-by-step walkthrough of the UI on our [Navigating the Kubecost UI](https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui) page, which will bring greater attention to Budgets (if we feel users do not fully recognize its utility and how to access it), along with all other main UI pages

thomasvn:
I really like this update, and feel like we should continue to do more of this! Thanks @chipzoller! It's great to have our technical docs point to our blog posts to expand on interesting use cases and examples.

Up to @bstuder99 to decide whether we should begin making more of an official home for these examples in the technical docs themselves.

chipzoller:
> I think it needs some expansion to exist as its own technical doc

What would you like to see here in order to be acceptable?

using-kubecost/navigating-the-kubecost-ui/budgets.md
```diff
@@ -60,3 +60,7 @@ Selecting _Edit_ in the row of a specific budget will open a window allowing you
 ### Deleting a budget
 
 Selecting _Delete_ will open the Delete Budget window. Confirm by selecting _Delete_.
+
+## Use cases
+
+For how to use budgets in a use case preventing cost overruns before they occur, see the guide [here](/using-kubecost/proactive-cost-controls.md).
```
kwombach12, line 66: wording a tad awkward. Since this is under 'use cases' i'd suggest simply saying: "Enforce Kubecost Budgets to prevent overruns before they occur" -- you could maybe make the whole use case link to the guide

kwombach12 (changes requested):

kwombach12:
> > I think it needs some expansion to exist as its own technical doc
> 
> What would you like to see here in order to be acceptable?

I _think_ what @bstuder99 is alluding to here is that it feels almost a bit backward to link from technical docs to a blog -- in general I feel like its typically the other way around. Not opposed to how we have it structured in this PR as I know the walkthrough in the blog is on the longer side.

using-kubecost/proactive-cost-controls.md
```diff
@@ -0,0 +1,13 @@
+# Proactive Cost Controls
+
+In addition to providing visibility into costs and how they are allocated, Kubecost is also able to take proactive steps to prevent some types of cost overruns before they occur.
+
+Kubernetes deployments and their pods consume resources on the underlying nodes on which they run. Kubecost is able to monitor this usage, down to the container level, and calculate how much a specific deployment out of many cost based on a variety of factors. This monitoring and subsequent reporting require that the deployment exist and be running for sometime and, therefore, cost some amount of money. Rather than allowing some deployments to be created and consume excess money, they can be blocked. For example, you as a Kubecost administrator may wish to restrict the `ticketing` namespace to a monthly allowable budget of $500. Once that budget has been consumed, you wish to prevent any further deployments into that namespace as these will cause a cost overrun.
```
kwombach12, line 5:
```
This monitoring and subsequent reporting require that the deployment exist and be running for sometime and, therefore, cost some amount of money.
``` 
Feels unnecessary

kwombach12, line 5: Is it just new deployments that we can block? Can't we also block against changes to existing deployments that would overrun a budget?

bstuder99:
Not necessarily, we've linked to blogs in the past to reference example use cases. I'm saying what is described in this doc should be more of a step-by-step tutorial or something more in-depth to constitute a tech doc, otherwise this content can just be added to the Budget doc to clarify the feature's purpose more

using-kubecost/proactive-cost-controls.md
```diff
@@ -0,0 +1,13 @@
...
+
+Using a combination of the [predictions API](/apis/governance-apis/spec-cost-prediction-api.md) and [budgets](/using-kubecost/navigating-the-kubecost-ui/budgets.md), Kubecost is able to solve this challenge with the addition of a dynamic admission controller such as [Gatekeeper](https://open-policy-agent.github.io/gatekeeper/website/) or [Kyverno](https://kyverno.io). In this process, a Kubernetes admission controller is installed in the cluster along with a policy which asks Kubecost for cost predictions and verifies those against a matching budget. The decision to allow or deny then rests with the admission controller.
```
kwombach12, line 7: instead of 'solve this challenge', I would suggest 'enforce budgets'

using-kubecost/proactive-cost-controls.md
```diff
@@ -0,0 +1,13 @@
...
+
+{% hint style="info" %}
+Note that currently budgets for the use of proactive controlling of costs should be confined to single-cluster installations if the workload grouping is anything other than "Cluster".
```
kwombach12, line 10: this is a bit confusing to me if I'm a new user... Also, I'd suggest we say 'Note:' instead of 'Note that'

kwombach12 (dismissed):

using-kubecost/proactive-cost-controls.md
chipzoller: Reworded.

using-kubecost/proactive-cost-controls.md
chipzoller: Changed.

using-kubecost/proactive-cost-controls.md
chipzoller: Simplified.

using-kubecost/navigating-the-kubecost-ui/budgets.md
chipzoller: Like this?

chipzoller (commented):
A few things here:

1. I didn't want to regurgitate the entire series of steps covered in the blog post in the same level of detail. It seems unnecessary to do so when the blog already does a decent job of this and in a step-by-step manner.
2. It isn't uncommon at all for technical documentation to reference a blog post (from the same project/product) as a form of auxillary documentation. Reference, usage, and use case docs are all different forms of docs and linking out to a different form admist another is common.
3. Budgets exist to service more than just this use case. Mixing use case docs (i.e., this one) with usage docs (the budgets docs) or reference docs (the API docs) generally doesn't work well since they have different goals.

Glad to make additional changes requested by the docs maintainers or, alternatively, if you feel this just isn't appropriate, feel free to close this PR entirely.

bstuder99:
It sounds like this is content that could be directly added into the Budgets/Prediction API doc; once the user has understood how Budgets works, they may wonder if there is a way to directly control consumption of the resource, which the basic feature can't do. Then we can link them to our blog post which contains a Kyverno tutorial, or write a separate tech doc for Gatekeeper. Ultimately I don't know if the current doc is really explaining a whole lot that the blog it links to doesn't. The note about single-cluster Kubecost installations is an important technical detail, but other than that, this reads like a blog post.

chipzoller (commented):
Added new headings with links to this doc.

chipzoller:
Thanks for the edits. Will you handle the merging or is that up to me?

srpomeroy:
Once approved, you can merge when ready.

PR #982, opened 2024-04-08
More details on how to tune the aggregator.

thomasvn:
## Related issue #

N/A

## Proposed Changes

Provides further description on what the env var does.


install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -88,14 +88,21 @@ kubecostAggregator:
     #
     # default: 10m
     DB_BUCKET_REFRESH_INTERVAL: 1h
-    # governs parallelism of derivation step
-    # more threads speeds derivation, but requires significantly more
+
+    # How much data to ingest from the federated store bucket, and how much data
+    # to keep in the DB before rolling the data off.
+    # default: 91
+    ETL_DAILY_STORE_DURATION_DAYS: "365"
```
michaelmdresser, line 95: I wonder if this higher recommended default might cause some pathology, see https://kubecost.slack.com/archives/GNS5G863C/p1712616956114309?thread_ts=1712588301.960419&cid=GNS5G863C

michaelmdresser (commented):
@kwombach12 this PR relates to my comment on the gating doc: https://docs.google.com/document/d/1Pa4IhDHftEcpHZIHbrztomIj3-3xNTkQ4oR7joqvO4k/edit?disco=AAABK_Gu36A

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: In most cases I've seen, Enterprise-scale users want approximately 1yr worth of data to look back on. Would it be helpful to add a message like:

> Note: If increasing this value to backfill historical data, it will take time to gradually ingest & process those historical ETL files.

It's worth noting that as of v2.2.1, the priority of ingestion should be 1) daily ETL sorted by most recent timestamps, then 2) hourly ETL sorted by most recent timestamps. So we work backwards in time.

michaelmdresser: That would be helpful to add, I just meant to say that increasing this without modifying `numDBCopyPartitions` and/or `DB_BUCKET_REFRESH_INTERVAL` may make data ingestion much slower than expected. Your suggested message remains valid even in that case.

michaelmdresser (dismissed):

thomasvn:
@bstuder99 I believe I've addressed Michael's comment above. This is ready for review & merge!

michaelmdresser (approved):

PR #983, opened 2024-04-09
Addresses a bug where Kubecost would report cost metrics in the billions

bstuder99:
## Related issue #

See https://github.com/kubecost/features-bugs/issues/60


srpomeroy:
Do we want to be enabling custom prices even for environment that are cloud-based where we support reconciliation?

michaelmdresser (dismissed):

michaelmdresser:
Oh, true, good point from spom -- answer: probably not

architecture/pricing-sources-matrix.md
```diff
@@ -60,6 +60,8 @@ kubecostProductConfigs:
     internetNetworkEgress: 0.12 # per GB per month cost
 ```
 
+Setting `customPricesEnabled` to `true` or enabling custom pricing through the UI has been shown to fix an error where Kubecost reported absurdly high cost metrics.
```
AjayTripathy, line 63: I don't think this is right; I think this should just be a line about rancher kubernetes engine clusters needing to have custom prices enabled.

bstuder99: If this only affects Rancher setups, we should just move this content into this article: https://docs.kubecost.com/install-and-configure/install/provider-installations/istio-rancher

AjayTripathy: afaik it's only rancher.

bstuder99:
I've removed the initial changes and added the content into a new Troubleshooting section for our Istio/Rancher install guide

PR #993, opened 2024-04-11
Adjusts singular Helm flag descriptions using config instead of `--set` commands

bstuder99:


install-and-configure/install/provider-installations/air-gapped.md
```diff
@@ -19,7 +19,7 @@ helm template kubecost --repo https://kubecost.github.io/cost-analyzer cost-anal
 ```
 
 {% hint style="info" %}
-The alpine/k8s image is not used in real deployments. It is only in the Helm chart for testing purposes.
+The alpine/K8s image is not used in real deployments. It is only in the Helm chart for testing purposes.
```
jessegoodier, line 22:
```suggestion
The alpine/k8s image is not used in real deployments. It is only in the Helm chart for testing purposes.
```

images are case sensitive, should be lower case

jessegoodier (approved):
Just commit the suggestion and this is good to go. thanks!

install-and-configure/install/provider-installations/air-gapped.md
bstuder99: Good call out

PR #994, opened 2024-04-11
Adds warning about disabling Prometheus

bstuder99:
## Related issue #

Based on user feedback that they did not realize they shouldn't disable Prometheus.

### DIFF
--- a/architecture/architecture.md
+++ b/architecture/architecture.md
@@ -19,7 +19,7 @@ Today, the core Kubecost product can be run with just components 1 and 2.1. See 
 ![Architecture Overview](/images/arch.png)
 
 {% hint style="warning" %}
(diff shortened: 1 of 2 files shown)


jessegoodier (approved):
Slightly reworded, thanks for this!

bstuder99:
@jessegoodier Are your commits accurate? I used our Prometheus Configuration Guide as a reference which claims that while Prom can be disabled (thus it technically is optional), but will result in 70-90% fewer costs returned. If that statement is inaccurate, I have to update that doc also.

jessegoodier:
> @jessegoodier Are your commits accurate? I used our Prometheus Configuration Guide as a reference which claims that while Prom can be disabled (thus it technically is optional), but will result in 70-90% fewer costs returned. If that statement is inaccurate, I have to update that doc also.

great question. 
There is an opencost use case for this, and technically kubecost could be configured the same. 
For support purposes, I would err on the side of best user experience and less confusion.

PR #996, opened 2024-04-12
Adds notices about device plugins

### LINKED ISSUE #995: [Enhancement] Add in docs that a device plugin is a requirement to track GPU costs in allocations
### Description

having a [device plug in](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/) a requirement to see gpuCost allocated in the allocations page. If not, users will only see it in the Assets page when drilling down into the GPU node. It won't get allocated to the workloads in allocations page for pods running on the same node and will read as 0.

bstuder99:
Closes #995 

## Proposed Changes

Adds warning banner to Allocation articles that explain how to view GPU costs.


mmurph3:
I might've spoke too soon. We are actually validating if this is in fact the case. Investigating if it's because no workloads are allocated to the gpu nodes or if the plug in is what we need. Will follow up shortly.

PR #998, opened 2024-04-15
Update aws-cloud-integration-using-irsa.md

kc-adawson:
Added method for using default kubecost service account instead of creating via eksctl. Both options are now highlighted in the document.


jessegoodier (approved):

bstuder99:
Let me go through with edits before we merge this

bstuder99 (approved):

Issue #960, opened 2024-04-01
[Enhancement] Add more troubleshooting steps for Kubecost 2.0


mmurph3:
### Description

Document more scenarios for troubleshooting

bstuder99:
I need you to be more concrete with this issue. Are you able to provide user feedback, or Support tickets related to v2.0 troubleshooting?

mmurph3:
Need to get more examples but lately we requested DB dumps for most scenarios where there's missing data, data isn't aligning with bill or derivation is failing. We may not need it in all these scenarios but maybe we could add a link directing users how to do a database dump: https://github.com/kubecost/kubecost-utilities/tree/main/aggregator-backup#reliable-usage when it is requested to start.

Another scenario: If data is missing, run the following queries to check whether the node's metadata is in the db. We could probably elaborate more here.

`kubectl exec -it KUBECOST-AGGREGATOR-POD-NAME sh`

# Go to the directory where the database is
```
cd /var/configs/waterfowl/duckdb/v0_9_2
ls -lah
```

# Copy the current database to a new file that we can work against
`cp kubecost-1718359086.duckdb.read kubecost-1718359086.duckdb.read.kubecost.copy`

# Open a DuckDB REPL pointed at the database copy
`duckdb kubecost-1718359086.duckdb.read.kubecost.copy`

Next, run the following queries to debug and check if node data is available:
```
show tables;
describe node_1h;
select * from node_1h;
select providerid,windowstart,windowend,* from node_1h;

.maxrows 100;
```

mmurph3:
1. This is to determine only if node metadata exists in the user's database as a means of troubleshooting the cause of missing data? This will show if there's hourly data for the node and includes the cluster name and window. Quick way to check if node data is making it to the db because if it's in the DB, it should show up in the UI.
2. What is kubectl exec -it KUBECOST-AGGREGATOR-POD-NAME sh doing in relation to the remaining tutorial?

This gets us a shell into aggregator pod so we can then get to the database.

3. See output below

```
D select providerid,windowstart,windowend,* from node_1h;
┌──────────────────────┬──────────────────────┬───┬──────────────────────┐
│      ProviderID      │     WindowStart      │ … │ OverheadCostFraction │
│       varchar        │ timestamp with tim…  │   │        double        │
├──────────────────────┼──────────────────────┼───┼──────────────────────┤
│ kind://docker/kind…  │ 2024-04-08 16:00:0…  │ … │                  0.0 │
[...]

PR #1000, opened 2024-04-15
Updates Aggregator doc with new troubleshooting method

bstuder99:


bstuder99:
Just need my questions in #960 resolved so we can finish this up

mmurph3 (approved):
lgtm

PR #1005, opened 2024-04-16
Fix CloudCost API doc to remove reference to `costMetric`

thomasvn:
## Related issue

N/A

## Proposed Changes

The CloudCost API doc references `costMetric` as an acceptable parameter. However according to the code ([ref](https://github.com/thomasvn/opencost/blob/531641e608f404bbdc756c5dd291a44367053190/pkg/cloudcost/queryservice_helper.go#L14-L65)) it is not actually parsed as a parameter.

I believe the `costMetric` parameter is used in the `/model/cloudCost/view` API however that is an internal undocumented API.

@Sean-Holcomb to confirm?


thomasvn:
Actually, this PR may not be as simple as removing `costMetric`. I'm realizing now that there's a whole section in the doc describing `costMetric`.

Sean-Holcomb:
While `costMetric` isn't a param for the basic `/cloudCost` endpoint, documentation describing it is still relevant because the return contains fields for each costMetric. The View endpoints take the `costMetric` because they only return one `costMetric`

bstuder99:
Gonna tackle this in a few and get it ready for publishing

bstuder99:
Ok, I took this PR and ran with it a bit but have made the following changes:
* Removed mention of costMetric param entirely from Cloud Cost API doc
* Updated filtering params in Cloud Cost API doc, updated Filter Params doc to reflect this also
* Updated the example payload of the Cloud Cost API

bstuder99:
Updating doc also to improve the example query endpoint and payload, will req review when done

apis/monitoring-apis/cloud-cost-api.md
```diff
@@ -31,85 +27,76 @@ Refers to the number of line items you are offsetting. Pairs with `limit`. See t
...
-| CostMetric value | Description                                                                                 |
-| ---------------- | ------------------------------------------------------------------------------------------- |
-| NetCost          | Costs inclusive of discounts and credits. Will also include one-time and recurring charges. |
-| AmortizedNetCost | `NetCost` with removed cash upfront fees and amortized                                      |
-| ListCost         | CSP pricing without any discounts                                                           |
-| InvoicedCost     | Pricing based on usage during billing period                                                |
-
-Providing a value for `CostMetric` is optional, but it will default to `AmortizedNetCost` if not otherwise provided.
-
-See our [Cloud Cost Metrics](/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer/cloud-cost-metrics.md) doc to learn more about these cost metric types and how they are calculated.
```
Sean-Holcomb, line 99: maybe keep this link with a change of text

Sean-Holcomb (approved):

apis/monitoring-apis/cloud-cost-api.md
bstuder99: This section exists almost verbatim in the Cloud Cost Explorer doc for the UI, which I think is where it will be more relevant.

PR #1006, opened 2024-04-17
New doc for Postgres Integration feature.

thomasvn:
## Related issue

Documentation for https://github.com/kubecost/cost-analyzer-helm-chart/pull/3238

## Proposed Changes

New doc for Postgres Integration feature.


integrations/postgres-integration.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Customized dashboarding used by FinOps/Analytics teams.
+- Single view populated by multiple data sources, to be used by developer/finance/operations teams.
+- Customized budgeting and forecasting via an in-house data pipeline.
+
+## Usage
+
+Prerequisites:
+
+- A running Postgres instance that is reachable by Kubecost.
```
cliffcolvin, line 19: I think we have to call out a requirement to create a user with the following permissions:
CREATE, INSERT, UPDATE

I'm not sure if we think the GRANT CREATE will cause any turbulence but if we keep the CREATE IF NOT EXISTS line in the integration it is a requirement.

thomasvn: Good catch. I've just added in a line describing this.

chipzoller: This may be out of scope for this PR, but what I'd ultimately like to see is maybe a quick start guide for this integration (strictly called out as non-production grade) using something like [Zalando's postgres operator](https://github.com/zalando/postgres-operator).

integrations/postgres-integration.md
```diff
@@ -0,0 +1,131 @@
...
+- Credentials for a Postgres user which has `CREATE`, `INSERT`, and `UPDATE` permissions on the database.
+
+### Step 1: Configure Helm values
+
+The below YAML is an example of how to configure the Postgres integration in your Helm values file. Notice that there are four different cronjobs being created (2 allocation queries, 1 asset query, 1 cloudCost query). Each cronjob is configured to run every `12h` and write to their respective tables in the Postgres database.
```
chipzoller, line 24: While the internal implementation here is a cronjob, I'd just state the user-apparent aspect which just scheduled runs. Minor point though.

integrations/postgres-integration.md
```diff
@@ -0,0 +1,131 @@
...
+# REQUIRED. A Kubecost Enterprise license key.
+kubecostProductConfigs:
+  productKey:
+    enabled: true
+    key: "my-enterprise-key-here"
+```
+
+Note, that multiple inserts into the database per day will not create duplicate data. All queries to the database are keyed on a unique `name`, `windowStart`, and `windowEnd`. If a key already exists in the database, we update the entry with the new data.
+
+You can also specify a k8s secret containing the database credentials via `databaseSecretName`. Below is an example of what the k8s secret should look like.
```
chipzoller, line 84:
```suggestion
You can also specify a Kubernetes secret containing the database credentials via `databaseSecretName`. Below is an example of what the secret should look like.
```

integrations/postgres-integration.md
```diff
@@ -0,0 +1,131 @@
...
+
+<details>
+<summary> Example k8s secret </summary>
```
chipzoller, line 87:
```suggestion
<summary> Example Kubernetes secret </summary>
```

chipzoller (changes requested):
Some very minor things and one suggestion. Treat the suggestion as optional for now, but I think it'd be a real nice-to-have for quick enablement.

bstuder99 (dismissed):

integrations/postgres-integration.md
chipzoller: Sure, that also works.

integrations/postgres-integration.md
thomasvn: I've updated the wording to say "query" instead of "cronjob". Does that work better?

bstuder99 (dismissed):

PR #1017, opened 2024-05-01
Update Agg docs with additional tuning guidance.

thomasvn:
## Related issue

- N/A

## Proposed Changes

- Edits the Aggregator doc, specifically the "Aggregator Optimizations" section. Adds additional info which was derived from the [Internal Aggregator Tuning Doc](https://docs.google.com/document/d/1CxGlr6GK8-OdEjgbStW4JrYSK1rJEao0ZIG5nw2Iiog). Specifically, pulls from the ~medium scale section.


install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -82,12 +82,12 @@ kubecostAggregator:
...
     #
-    # Note, that the default value is set to 10m to optimize for the
+    # Note, that the default value is set to 60m to optimize for the
     # first-install experience of Kubecost (i.e. it prioritizes small data
     # becoming available more quickly).
     #
-    # default: 10m
-    DB_BUCKET_REFRESH_INTERVAL: 1h
+    # default: 60m
+    DB_BUCKET_REFRESH_INTERVAL: 2h
```
ameijer, line 90: I'm on the fence about whether to include this. With our new tuning logic for this variable in 2.3, this variable becomes even more of a foot gun than it already is

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -117,10 +137,10 @@ kubecostAggregator:
     storageRequest: 512Gi
   resources:
     requests:
-      cpu: 1000m
-      memory: 1Gi
+      cpu: 4
+      memory: 12Gi
     limits:
-      # cpu: 2000m
+      cpu: 6
```
ameijer, line 143: I wonder if re-structiring this document to take the same small/medium/large/extra large set of recommendations as in the aggregator tuning google doc would be a better approach? for example, a user with 3 nodes and 150 allocations a day will way overprovision this and kubecost will seem expensive. 

The $1,000,000 question is what is a "larger aggregator deployment"? I wonder if we establish a rough mapping, and ask users to check their container count via the new diagnostic endpoint in the UI. These settings would be good for maybe 20-60k containers per day. Below 10k allocations per day, the prior settings would probably work

ameijer (commented):
some discussion starter thoughts!

bstuder99 (approved):

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: Right, given that v2.3 will now dynamically adjust this variable, maybe we remove it from recommended configurations? I do still think we should explain what this env var is in the troubleshooting section?

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: I'm in agreement with all the above! I'm thinking we just start with details for small/medium environments as that should cover a majority of cases for now.

I'll update this doc to establish a rough mapping for what configs to use based on the number of containers per day. I'll err on the side of overprovisioning rather than underprovisioning.

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: Gave it a bit more thought, and I think I'm going to leave the recommendation of `DB_BUCKET_REFRESH_INTERVAL` as is. In most cases this is the sweet spot for Enterprise-scale users. And v2.3's ability to dynamically adjust this variable should move it into a more appropriate range if needed.

If this becomes a footgun, let's remove this from the doc in future iterations!

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -59,19 +63,18 @@ serviceAccount:
 
 ### Aggregator Optimizations
 
-For larger deployments of Kubecost, Aggregator can be tuned.
+For larger deployments of Kubecost, Aggregator can be tuned. The settings below are in addition to the basic configuration above.
+
+{% hint style="info" %}
+This configuration is estimated to be sufficient for environments monitoring < 60k unique containers per day. You can check this metric on the `/diagnostics` page.
+{% endhint %}
```
thomasvn, line 70: @bstuder99 @ameijer Added this line to give users an estimate for what scale these configurations should be sufficient for. We'll need to continuously update this as we learn more from users. I added a similar line to the "Basic configuration" section.

If we're good with this, we can approve & merge the PR!

ameijer (approved):

PR #1022, opened 2024-05-01
AWS troubleshooting fixes

### LINKED ISSUE #1018: [Enhancement] Add note/warning about potential need to increase Lambda concurrent execution quota
### Description

In step 5 of integrating Kubecost with AWS [here](https://docs.kubecost.com/install-and-configure/install/cloud-integration/aws-cloud-integrations#step-5-provide-cur-config-values-to-kubecost), there is a chance some users will run into failure when running the CFT in their account due to low Lambda concurrent execution values. The fix for this is to request a quota increase in the AWS console. By default, the value may be set at 10, which is too low to allow the template to execute. Further details can be found [here](https://benellis.cloud/my-lambda-concurrency-applied-quota-is-only-10-but-why).

image

We need to add a note/warning in this relevant section of the docs explaining about this and directing users to perform a quota increase.

### LINKED ISSUE #1020: [Enhancement] Making external AWS docs on Athena setup mandatory
### Description

In step 2 of the AWS integration [here](https://docs.kubecost.com/install-and-configure/install/cloud-integration/aws-cloud-integrations#step-2-setting-up-athena), the first paragraph currently shows the following.

> As part of the CUR creation process, Amazon also creates a CloudFormation template that is used to create the Athena integration. It is created in the CUR S3 bucket, listed in the Objects tab in the path s3-path-prefix/cur-name and typically has the filename crawler-cfn.yml. This .yml is your necessary CloudFormation template. You will need it in order to complete the CUR Athena integration. For more information, see the AWS doc [Setting up Athena using AWS CloudFormation templates](https://docs.aws.amazon.com/cur/latest/userguide/use-athena-cf.html).

The last sentence there suggests that the link to the AWS page is optional or informative. It is not. This step MUST be followed in order to proceed to additional steps.

The docs need to clearly state users should follow the linked procedure before proceeding to the next step or else the integration will not be possible.

bstuder99:
Closes #1018 
Closes #1020


bstuder99:
Ignore link checker, picking up false positive

chipzoller (approved):
Nice turnaround on these.

PR #1024, opened 2024-05-02
addresses zd-6287, where user was seeing discrepancies between amortized and net amortized costs

williamkubecost:
## Related issue # 

Closes https://kubecost.zendesk.com/agent/tickets/6287

## Proposed Changes

Needed to update docs to reflect proper definitions for amortized costs and net amortized costs as per opencosts [codebase](https://github.com/opencost/opencost/blob/194ebd1cbb2288002af690b75da5d0cbc26905c8/pkg/cloud/gcp/bigqueryintegration_types.go#L192)

### DIFF
--- a/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer/cloud-cost-metrics.md
+++ b/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer/cloud-cost-metrics.md
@@ -110,7 +110,8 @@ The Cost column plus the sum of all credit amounts.
 
 **Amortized Net Cost**
 
-Kubecost uses Net Cost.
+Amortized Net Cost is Cost with all credits and amortized CUD payments
+`amortizedNetCost := cost + creditAmount + cudCreditAmount + flexibleCUDCreditAmount + flexibleCUDNetPayedAmount`
 
 **Invoiced Cost**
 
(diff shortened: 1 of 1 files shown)


using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer/cloud-cost-metrics.md
```diff
@@ -110,15 +110,17 @@ The Cost column plus the sum of all credit amounts.
 
 **Amortized Net Cost**
 
-Kubecost uses Net Cost.
+Amortized Net Cost is Cost with all credits and amortized CUD payments
+`amortizedNetCost := cost + creditAmount + cudCreditAmount + flexibleCUDCreditAmount + flexibleCUDNetPayedAmount`
```
biancaburtoiu, line 114: Since the cloud cost pipeline code is open source, we could add a reference pointing directly to the source code for users that are interested in the in-depth implementation details. I think the doc can be refactored to add the notes on how each cost type is computed, but I don't think we need to show the actual equations. That's just a personal opinion though, deferring to @bstuder99 here!

bstuder99: I think both is always good if we can provide it, I like the addition of the source code

williamkubecost: I think having the equations is good, personally it makes it easier to understand the math.

PR #1031, opened 2024-05-07
NetworkCosts config update.

thomasvn:
## Proposed Changes

* Remove reference to `.config.services` since this will now be on by default in `v2.3`. https://github.com/kubecost/cost-analyzer-helm-chart/pull/3383
* Cleanup section about "Log Level" since this configuration is now supported via Helm. https://github.com/kubecost/cost-analyzer-helm-chart/pull/3278

### DIFF
--- a/install-and-configure/advanced-configuration/network-costs-configuration.md
+++ b/install-and-configure/advanced-configuration/network-costs-configuration.md
@@ -30,7 +30,7 @@ There are Grafana dashboards that are included with the Kubecost installation, b
 
 ## Enabling network costs
 
-To enable this feature, set the following parameters in your values file following the [Helm installation](https://kubecost.com/install) guide.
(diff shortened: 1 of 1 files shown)


thomasvn:
TODO. Now that we are turning the below configs on by default, we should consider providing guidance on when these configs can be safely turned off. And what the implications are. Maybe document directly in the `values.yaml`.

```yaml
  config:
    services:
      # google-cloud-services: when set to true, enables labeling traffic metrics with google cloud
      # service endpoints
      google-cloud-services: false
      # amazon-web-services: when set to true, enables labeling traffic metrics with amazon web service
      # endpoints.
      amazon-web-services: false
      # azure-cloud-services: when set to true, enables labeling traffic metrics with azure cloud service
      # endpoints
      azure-cloud-services: false
```

bstuder99:
Thomas I was going to merge this into the v2.3 branch, but can you just let me know why log level is no longer configurable?

PR #1038, opened 2024-05-14
Anomaly Detection updates

bstuder99:


kwombach12:
@bstuder99 everything you have looks good - would just recommend we add a brief explanation/walkthrough of the 'drill in' experience

bstuder99:
@kwombach12 Can you elaborate?

kwombach12:
> @kwombach12 Can you elaborate?

I think its worth highlighting and showing the user experience when you click on one of the rows in the anomaly detection dashboard. I.e. when you click on a cloud cost anomaly, you are brought to the cloud costs dashboard, pre filtered, with the lookback window and anomalous day highlighted 
Screenshot 2024-05-15 at 12 37 55 PM

bstuder99:
Ah ok. I've been testing this feature and sometimes do not understand why some of these results, like your pic, are considered anomalies. We see spend decrease through the lookback window, and the highlighted day is only after several days of the same cost. Is it comparing the average cost per day within the lookback window to the very next step interval?

kwombach12:
Yes - it looks at the average spend in the lookback window and then looks if the 'current day' is an outlier from that mean

chipzoller:
@kwombach12, same question.

kwombach12:
> @kwombach12, same question.

yea lets close this pr

PR #1040, opened 2024-05-16
Updates Thanos Aggregator migration diagrams 

bstuder99:
## Proposed Changes

Updated architecture diagram using draw.io for a cleaner and more legible look. Just a basic draft, if we want to tweak any of the design choices (colors etc), let me know


thomasvn (approved):
Thanks for doing this @bstuder99! Is there any place you're editing/saving your diagrams at? Would love for it to be a shared location so we can continuously iterate & collaborate on them.

bstuder99:
We should have a shared Google Drive folder through draw.io where these are saved, let me know if you can't find them

PR #1043, opened 2024-05-22
Updates more arch diagrams and updates federated ETL description

bstuder99:
I updated several of our architecture diagrams to look more visually similar and easier to understand. I also updated an outdated diagram for Federated ETL and updated to corresponding doc to match the description


images/diagrams/fed-etl-agg-arch.png
thomasvn, line 1: Could we simplify "Queried federated data"? Maybe replace with "Multicluster Kubecost data"? Really, it just needs to encapsulate the fact that both your UI and your API calls will now return data about all your monitored clusters.

thomasvn:
@bstuder99 Looking good! Thanks for your work standardizing these diagrams, and also updating their corresponding text.

images/diagrams/fed-etl-agg-arch.png
bstuder99: Yeah thats a better way of wording it. Was mainly trying to adhere to wording from the original diagram but I like this suggestion

chipzoller:
@thomasvn, are we good to merge this into `main` at this time?

thomasvn (approved):
Yes, let's go ahead and merge!

PR #1046, opened 2024-05-29
Updated recommendations for deploying Kubecost on secondaries.

thomasvn:
## Proposed Changes

As titled

### DIFF
--- a/install-and-configure/install/multi-cluster/secondary-clusters.md
+++ b/install-and-configure/install/multi-cluster/secondary-clusters.md
@@ -12,11 +12,11 @@ See the [Additional resources](/install-and-configure/install/multi-cluster/seco
 
 ## Kubecost global
 
-Disable product caching and reduce query concurrency with the following parameters:
+Disable unnecessary containers/pods on secondary clusters.
 
 ```
---set federatedETL.federatedCluster=true
 --set federatedETL.agentOnly=true
+--set federatedETL.federatedCluster=true
 ```
 
 ## Grafana
@@ -32,15 +32,9 @@ Grafana is not needed on secondary clusters.
 
 Kubecost and its accompanying Prometheus collect a reduced set of metrics that allow for lower resource/storage usage than a standard Prometheus deployment.
 
-The following configuration options further reduce resource consumption when not using the Kubecost frontend:
-
 ```
 --set prometheus.server.retention=2d
 ```
-
-Potentially reducing retention even further, metrics are sent to the storage-bucket every 2 hours.
-
-You can tune `prometheus.server.persistentVolume.size` depending on scale, or outright disable persistent storage.
 
 ## Node-Exporter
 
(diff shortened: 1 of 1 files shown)


install-and-configure/install/multi-cluster/secondary-clusters.md
```diff
@@ -32,16 +32,10 @@ Grafana is not needed on secondary clusters.
 
 Kubecost and its accompanying Prometheus collect a reduced set of metrics that allow for lower resource/storage usage than a standard Prometheus deployment.
 
-The following configuration options further reduce resource consumption when not using the Kubecost frontend:
-
 ```
 --set prometheus.server.retention=2d
```
srpomeroy, line 36: Is there a reason we should leave this helm value here?

thomasvn: Although the default Prometheus retention is now 97h ([code ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/b128331ceaabb44e5bc4b2b1d3fd3dbe4abca52e/cost-analyzer/values.yaml#L1269-L1294)), I think it's still worth pointing out that this can be further tuned if needed! What do you think?

install-and-configure/install/multi-cluster/secondary-clusters.md
```diff
@@ -12,7 +12,7 @@ See the [Additional resources](/install-and-configure/install/multi-cluster/seco
 
 ## Kubecost global
```
chipzoller, line 13: I know you didn't modify this, but what does "Kubecost global" mean here? The primary cluster?

install-and-configure/install/multi-cluster/secondary-clusters.md
```diff
@@ -12,7 +12,7 @@ See the [Additional resources](/install-and-configure/install/multi-cluster/seco
...
 
-Disable product caching and reduce query concurrency with the following parameters:
+Disable unnecessary containers/pods on secondary clusters.
```
chipzoller, line 15: Are you recommending users to disable Kubecost containers/Pods here or other workloads? If the latter, why is this a suggestion? The verb "disable" in the context of Kubernetes also doesn't mean a whole lot; Pods are atomic and either exist or not. There is no present-but-disabled state. A nit, I realize, but as a user when I read this I don't quite understand.

install-and-configure/install/multi-cluster/secondary-clusters.md
chipzoller: The default retention value is now `97h` (~4 days) so this recommendation is a reduction of half.

install-and-configure/install/multi-cluster/secondary-clusters.md
thomasvn: My interpretation, is that it refers to components of Kubecost that are certainly running. Whereas sections below such as Grafana may not be running in user environments.

install-and-configure/install/multi-cluster/secondary-clusters.md
thomasvn: Good point about the verb "disable". But yes the interpretation is correct! We do not want to run pods such as forecating, grafana. And we do not want to run the cost-analyzer-frontend container. Thoughts on how to rephrase?

install-and-configure/install/multi-cluster/secondary-clusters.md
chipzoller: Would another way to phrase this be "Kubecost Primary Components"? Otherwise, I'm still not sure I understand as a reader.

install-and-configure/install/multi-cluster/secondary-clusters.md
chipzoller: I suggest we be prescriptive since we're the experts. Which Pods/containers can we tell users to disable? And how would they go about doing that? This is one valid example, I think, where a Helm values snippet is warranted in which we provide the prescription for a minimal secondary cluster configuration.

install-and-configure/install/multi-cluster/secondary-clusters.md
thomasvn: I've renamed the header to "Kubecost". I was hesitant to use the word "Primary Components" because this doc is specifically about secondary clusters. I also think just using the header "Kubecost" may be fitting because other headers just reference the app names (e.g. Prometheus, Grafana, Node-exporter). Let me know what you think!

install-and-configure/install/multi-cluster/secondary-clusters.md
thomasvn: In the code snippet below this text, we tell the user to set `--set federatedETL.agentOnly=true`. This is a Helm flag which when true will not render any of those extra pods/containers I listed above. Let me know what you think! Does something need to be more explicit?

chipzoller:
Comments all resolved, you feel?

thomasvn:
Yes I think so!

chipzoller (approved):

PR #1074, opened 2024-06-21
aggregator 2.3 tuning

jessegoodier:
Updating for 2.3

### DIFF
--- a/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
+++ b/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
@@ -88,6 +88,7 @@ kubecostAggregator:
   # `deployMethod: "statefulset"`
   replicas: 1
 
+
   # stagingEmptyDirSizeLimit changes how large the "staging"
   # /var/configs/waterfowl emptyDir is. It only takes effect in StatefulSet
   # configurations of Aggregator, other configurations are unaffected.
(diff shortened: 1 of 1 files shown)


chipzoller (approved):

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -77,74 +69,97 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
-    # first-install experience of Kubecost (i.e. it prioritizes small data
-    # becoming available more quickly).
-    #
-    # default: 60m
-    DB_BUCKET_REFRESH_INTERVAL: 2h
+  # deployMethod determines how Aggregator is deployed. Current options are
+  # "singlepod" (within cost-analyzer Pod) "statefulset" (separate
+  # StatefulSet), and "disabled". Only use "disabled" if this is a secondary
+  # Federated ETL cluster which does not need to answer queries.
+  deployMethod: singlepod
```
thomasvn, line 76: @jessegoodier I see the Aggregator Optimizations as a section mostly for Enterprise users. So we should probably recommend `deployMethod: statefulset` here?

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -77,74 +69,97 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
+  # fullImageName:
+  imagePullPolicy: IfNotPresent
+
+  # For legacy configuration support, `enabled: true` overrides deployMethod
+  # and causes `deployMethod: "statefulset"`
+  enabled: false
+
+  # Replicas sets the number of Aggregator replicas. It only has an effect if
+  # `deployMethod: "statefulset"`
+  replicas: 1
```
thomasvn, line 89: @jessegoodier In fact we should probably remove this whole section because these are already the defaults in the `values.yaml` file?

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -77,74 +69,97 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
+  # Allocation file with size 300MiB, this value should be set to approximately
+  # 2400Mi. In most environments, the default should suffice.
+  stagingEmptyDirSizeLimit: 2Gi
+
+  # this is the number of partitions the datastore is split into for copying
+  # the higher this number, the lower the ram usage but the longer it takes for
+  # new data to show in the kubecost UI
+  # set to 0 for max partitioning (minimum possible ram usage, but the slowest)
+  # the default of 25 is sufficient for 95%+ of users. This should only be modified
+  # after consulting with Kubecost's support team
```
thomasvn, line 107: @jessegoodier I think this comment may no longer be relevant? I think I remember you found that `0` won't work for this config. And from what I remember setting it to `1` will allow for a faster copy, but more memory usage.

jessegoodier: I'd like to get a full eng review on tuning and recommendations. 
agree the statefulSet should set here and remove defaults that will rarely, if ever be modified.

thomasvn: Yep sounds good. I will propose a new PR sometime soon and let's get a full eng review on it.

PR #1076, opened 2024-06-21
Update Aggregator tuning for v2.3

thomasvn:
## Related Issue

Builds upon https://github.com/kubecost/docs/pull/1074

## Proposed Changes

- Removes default values which do not need to be optimized
- Updated comments around what some of the configs do

### DIFF
--- a/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
+++ b/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
@@ -69,25 +69,55 @@ Because the Aggregator PV is relatively small, the least expensive performance g
 
 ```yaml
 kubecostAggregator:
-  # deployMethod determines how Aggregator is deployed. Current options are
-  # "singlepod" (within cost-analyzer Pod) "statefulset" (separate
-  # StatefulSet), and "disabled". Only use "disabled" if this is a secondary
-  # Federated ETL cluster which does not need to answer queries.
-  deployMethod: singlepod
+  logLevel: info
 
-  # fullImageName overrides the default image construction logic. The exact
-  # image provided (registry, image, tag) will be used for aggregator.
-  # fullImageName:
-  imagePullPolicy: IfNotPresent
+  # How much data to ingest from the federated store bucket, and how much data
+  # to keep in the DB before rolling the data off.
+  # 
+  # Note: If increasing this value to backfill historical data, it will take
+  # time to gradually ingest & process those historical ETL files. Consider
+  # also increasing the resources available to the aggregator as well as the
+  # dbConcurrentIngestionCount.
+  # 
+  # default: 91
+  etlDailyStoreDurationDays: 375
 
-  # For legacy configuration support, `enabled: true` overrides deployMethod
-  # and causes `deployMethod: "statefulset"`
-  enabled: false
+  # How many threads the read database is configured with (i.e. Kubecost API /
+  # UI queries). If increasing this value, it is recommended to increase the
+  # aggregator's memory requests & limits.
+  # default: 1
+  dbReadThreads: 1
 
-  # Replicas sets the number of Aggregator replicas. It only has an effect if
-  # `deployMethod: "statefulset"`
-  replicas: 1
+  # How many threads the write database is configured with (i.e. ingestion of
+  # new data from S3). If increasing this value, it is recommended to increase
+  # the aggregator's memory requests & limits.
+  # default: 1
+  dbWriteThreads: 1
 
+  # How many threads to use when ingesting Asset/Allocation/CloudCost data
+  # from the federated store bucket. In most cases the default is sufficient,
+  # but can be increased if trying to backfill historical data.
+  # default: 1
+  dbConcurrentIngestionCount: 1
+
+  # Memory limit applied to read database connections.
+  # default: 0Gi is no limit
+  dbMemoryLimit: 0Gi
+
+  # Memory limit applied to write database connections.
+  # default: 0Gi is no limit
+  dbWriteMemoryLimit: 0Gi
+
+  # If "true" can improve the time it takes to copy the write DB, at the expense
+  # of additional memory usage.
+  # default: "false"
+  dbCopyFull: "false"
+
+  # The number of partitions the datastore is split into for copying. The higher
+  # this number, the lower the RAM usage but the longer it takes for new data to
+  # show in the Kubecost UI.
+  # default: 1
+  numDBCopyPartitions: 1
 
   # stagingEmptyDirSizeLimit changes how large the "staging"
   # /var/configs/waterfowl emptyDir is. It only takes effect in StatefulSet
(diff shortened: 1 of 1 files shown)


thomasvn:
FYI @srpomeroy @mclean0328 @williamkubecost

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -69,88 +69,69 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
-  # fullImageName:
-  imagePullPolicy: IfNotPresent
-
-  # For legacy configuration support, `enabled: true` overrides deployMethod
-  # and causes `deployMethod: "statefulset"`
-  enabled: false
-
-  # Replicas sets the number of Aggregator replicas. It only has an effect if
-  # `deployMethod: "statefulset"`
-  replicas: 1
```
thomasvn, line 89: These are default values which don't need to be included in the "optimization" section of the doc.

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -69,88 +69,69 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
+  # How much data to ingest from the federated store bucket, and how much data
+  # to keep in the DB before rolling the data off.
+  # 
+  # Note: If increasing this value to backfill historical data, it will take
+  # time to gradually ingest & process those historical ETL files. Consider
+  # also increasing the resources available to the aggregator as well as the
+  # dbConcurrentIngestionCount.
+  # 
+  # default: 91
+  etlDailyStoreDurationDays: 375
```
thomasvn, line 83: In most enterprise use cases they will want a year of data to be retained in the DB.

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -69,88 +69,69 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
+  # How much data to ingest from the federated store bucket, and how much data
+  # to keep in the DB before rolling the data off.
+  # 
+  # Note: If increasing this value to backfill historical data, it will take
+  # time to gradually ingest & process those historical ETL files. Consider
+  # also increasing the resources available to the aggregator as well as the
+  # dbConcurrentIngestionCount.
+  # 
+  # default: 91
+  etlDailyStoreDurationDays: 365
```
thomasvn, line 83: Most enterprise use cases will want a year of data in the db.

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -69,88 +69,69 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
-  # also increasing the resources available to the aggregator as well as the
-  # refresh & concurrency env vars.
-  # 
-  # default: 91
-  etlDayStoreDurationDays: 91
+
+  # If "true" can improve the time it takes to copy the write DB, at the expense
+  # of additional memory usage.
+  # default: "false"
+  dbCopyFull: "false"
```
thomasvn, line 114: @mbolt35 Based on my reading of the code, setting this to `true` would create a separate DB Transaction when copying each chunk/partition of the table. This is probably a safer and less memory intensive approach. But probably also slower?

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -69,88 +69,69 @@ Because the Aggregator PV is relatively small, the least expensive performance g
...
+
+  # The number of partitions the datastore is split into for copying. The higher
+  # this number, the lower the RAM usage but the longer it takes for new data to
+  # show in the Kubecost UI.
+  # default: 1
+  numDBCopyPartitions: 1
```
thomasvn, line 120: @mbolt35 Similarly, reading this seemed to indicate that increasing the number of partitions/chunks would lead to higher volume of smaller sized INSERTS. Probably better for memory but takes longer to execute? Let me know if you agree/disagree.

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
mbolt35: So, this was added as sort of a quick test, but I actually think it might be valuable going forward to have this option. I think this description is probably "ok" for now, although I'm not sure how _true_ it actually is. 

Both methods are _very_ similar, and my guess is that they likely have a fairly equivalent memory footprint, but it's actually possible that the non-full-copy option (`false`) will use more memory due to potential parallelization that can happen via db connections (likely depends on the db write threads > 1). 

To also provide some clear context on _why_ this option was created (hint: it wasn't added for memory/speed tradeoffs). We were running into some contention issues _during_ the copy, so this change was actually attempting to create a more atomic copy operation such that the contending reads on the old db didn't explode the copy (or vice-versa). 

For reference, here's the "full copy": 
https://github.com/kubecost/kubecost-cost-model/blob/305e43fd85a27867a7ca677c7a636283a2498204/pkg/duckdb/write/writer.go#L720-L953

It's a quick copy/paste job that just appends all the commands into a single SQL string and is issued over 1 db connection inside a txn. 

The default is actually a bit more "up in the air" in terms of the execution semantics (variable based on inputs):
https://github.com/kubecost/kubecost-cost-model/blob/305e43fd85a27867a7ca677c7a636283a2498204/pkg/duckdb/write/writer.go#L444-L718

The "chunking" in this case is sort of hand-wavy, as the window copies run serially, and the memory pool for the database will only cleanup when the full copy is complete. 

In conclusion here, I think we still want to look into:
* Is this option necessary at all, or can we find a very optimal copy mechanism for most/all configurations
* Are there even major differences between the two, and if so, what are they? (to be documented here). 

The contention issues have mostly been resolved (by completely avoiding reads on the write db during copy).

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
mbolt35: From response above: 

> The "chunking" in this case is sort of hand-wavy, as the window copies run serially, and the memory pool for the database will only cleanup when the full copy is complete.

jessegoodier:
@thomasvn you good to merge this soon?

jessegoodier (approved):

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
thomasvn: Ok that context is really helpful. TLDR ... "it depends". Let's remove `dbCopyFull` and `numDBCopyPartitions` from docs and Helm defaults for now? We'll tune these as needed from a support perspective? @jessegoodier thoughts?

jessegoodier: I'm fine leaving in variables that we may want to be able to easily have customers change in the helm chart.
The comment already says:
# the below settings should only be modified with support from Kubecost staff

thomasvn: @jessegoodier Ok I'm fine with approving and merging this PR to ensure docs are in alignment with the v2.3.1 release. However we need to return to this section of the docs and modify/update it as we learn more about how these tuning variables holistically affect Aggregator performance.

jessegoodier (approved):

PR #1077, opened 2024-06-21
Troubleshooting ETL Utils

thomasvn:
## Related Issue

N/A

## Proposed Changes

- As titled


williamkubecost (dismissed):
lgtm

install-and-configure/install/multi-cluster/federated-etl/thanos-migration-guide.md
```diff
@@ -194,3 +194,25 @@ helm upgrade kubecost cost-analyzer --repo https://kubecost.github.io/cost-analy
 ## Troubeshooting
 
 To help diagnose problems with Aggregator, check the Aggregator container logs for query failures or SQL table failures. If you have additional questions, contact Kubecost support at [support@kubecost.com](mailto:support@kubecost.com).
+
+Additionally, reference the below additional information.
```
chipzoller, line 198:
```suggestion
See additional troubleshooting procedures below.
```

install-and-configure/install/multi-cluster/federated-etl/thanos-migration-guide.md
```diff
@@ -194,3 +194,25 @@ helm upgrade kubecost cost-analyzer --repo https://kubecost.github.io/cost-analy
...
+    # to process all files in the `/etl` directory of your bucket.
+    requests:
+      memory: 10Gi
+  # Use the most recent version of the Kubecost image. For example:
+  # "gcr.io/kubecost1/cost-model:prod-2.3.0"
+  fullImageName: gcr.io/kubecost1/cost-model:prod-x.x.x
+  env:
+    LOG_LEVEL: debug
+    # Set to 1 to use less memory. Default is 2.
+    ETL_UTIL_THREADS: 1
```
chipzoller, line 217: Which of these values are actually pertinent to resolving this OutOfMemory condition? I assume it's just the `etlUtils.resources.requests.memory` value, but this more explanation aside from just a chunk of YAML. Pretty sure we don't want to recommend users run with debug log level either.

chipzoller (changes requested):

install-and-configure/install/multi-cluster/federated-etl/thanos-migration-guide.md
thomasvn: Actually, the most pertinent value may be `fullImageName`. Because many users follow this migration guide while on v1.108, that would by default have the users running a v1.108 version of ETL Utils. I've found that more recent versions of ETL Utils to be better at managing its memory.

Also, `LOG_LEVEL: debug` is actually the default for ETLUtils since the command is not very verbose to begin with.

Aside from the comments left in YAML, do you think I need to write a few sentences about some of this to be more explicit?

chipzoller: Confused. Why/when would `fullImageName` be used here?

chipzoller: Ok, this just seems odd to me and I can see being potentially confusing as well as concerning to a customer/user. Do you think this is something we even should document for self consumption or just "contact support"?

thomasvn: If this is too niche/confusing a subject we can definitely find a way to internally document instead of externally document. As this is more a product decision, I'll let you make final call! Feel free to close this if you feel we should internally document instead.

chipzoller (approved):

PR #1085, opened 2024-07-02
update old aws tagging instructions

kwombach12:
move 'viewing account level tags' to main aws-cloud integrations page

### DIFF
--- a/install-and-configure/install/cloud-integration/aws-cloud-integrations/aws-cloud-integrations.md
+++ b/install-and-configure/install/cloud-integration/aws-cloud-integrations/aws-cloud-integrations.md
@@ -631,6 +631,16 @@ For AWS Lambda users, you may experience errors running the CloudFormation templ
 
 ![AWS Lambda](/images/aws-lambda.png)
 
+## Viewing account-level tags
+
+Account-level tags are applied (as labels) to all the Assets built from resources defined under a given AWS account. You can filter AWS resources in the Kubecost Assets View (or API) by account-level tags by adding them ('tag:value') in the Label/Tag filter.
+
+If a resource has a label with the same name as an account-level tag, the resource label value will take precedence.
+
+Modifications incurred on account-level tags may take several hours to update on Kubecost.
+
+Your AWS account will need to support the `organizations:ListAccounts` and `organizations:ListTagsForResource` policies to benefit from this feature.
+
 ## Summary and pricing
 
 AWS services used here are:


install-and-configure/install/cloud-integration/aws-cloud-integrations/aws-out-of-cluster.md
```diff
@@ -1,72 +0,0 @@
...
-| Cluster | kubernetes_cluster	| cluster-name	|
-| Namespace | kubernetes_namespace	| namespace-name |
-| Deployment | kubernetes_deployment	| deployment-name |
-| Label | kubernetes\_label\_NAME* | label-value    |
-| DaemonSet | kubernetes_daemonset	| daemonset-name |
-| Pod | kubernetes_pod	    | pod-name     |
-| Container | kubernetes_container	| container-name |
-
-{% hint style="info" %}
-In the `kubernetes_label_NAME` tag key, the `NAME` portion should appear exactly as the tag appears inside of Kubernetes. For example, for the tag `app.kubernetes.io/name`, this tag key would appear as `kubernetes_label_app.kubernetes.io/name`.
```
thomasvn, line 28: I think this detail is still important/relevant?

kwombach12: @thomasvn do you know where this info is relevant? My understanding is that these tagging instructions were only really relevant for 'external costs' (the removed column in allocations page)

thomasvn: @kwombach12 Ah that's right. I was thinking back to Advanced Reports as well. Do Collections need this level of tagging?

kwombach12: @thomasvn no -- collections leverages the same filtering as Cloud Costs so this tagging doesn't really enhance any existing features of our product as far as I can tell. I've also seen a couple cases where users get confused by these docs -- they do all this tagging and expect kubecost to allocate/handle that mapping and we don't anymore

thomasvn: Right, makes sense. Thanks for clarifying! I agree we should remove this section of the docs then

thomasvn (dismissed):

srpomeroy (approved):
This looks good.

We should probably update [this redirect](https://github.com/kubecost/docs/blob/f582ccd01edc7f539fefded4118a09eeb7a2e8f7/.gitbook.yaml#L9) to point to a different page.

This is a good alternative: `install-and-configure/install/cloud-integration/aws-cloud-integrations/aws-cloud-integrations.md`

kwombach12:
@srpomeroy thanks for that callout, just updated that redirect.

thomasvn (approved):

PR #1101, opened 2024-07-29
Remove DCGM_FI_DEV_NAME from example output

thomasvn:
## Related Issue

N/A

## Proposed Changes

As titled! This metric was not included in the list of "Kubecost custom metrics", so no other docs changes will need to be made.


chipzoller (approved):
Thanks!

PR #1102, opened 2024-07-30
FedETL Migration Guide does not require `DB_BUCKET_REFRESH_INTERVAL`

thomasvn:
## Related Issue

N/A

## Proposed Changes

- As titled. As of v2.3 this is no longer an environment variable that can be tuned.


MikeFindsThings (approved):
LGTM!

PR #1106, opened 2024-08-05
Fix iam policy file path

augustkang:
## Proposed Changes

Corrected the path for IAM policy files in the documentation from /aws/iam-policies-cur to /aws/iam-policies/cur to ensure accurate referencing.


thomasvn (approved):
Thanks!

PR #1109, opened 2024-08-06
ClaimValue entered into EntraID needs to match ClaimValue supplied via Helm config.

thomasvn:
## Related Issue

N/A

## Proposed Changes

- Both strings need to be matching. In the case of this doc, both need to be "admins".


mmurph3 (approved):
LGTM ! Thanks @thomasvn

PR #1110, opened 2024-08-14
clarify what clusters to enable annotation emission on

mmurph3:
## Proposed Changes
Clarified this needs to be done on each cluster user wants annotation emission from.


srpomeroy (approved):

thomasvn:
Nice, thanks @mmurph3!

PR #1120, opened 2024-08-29
docs : fix logic to access Kubecost UI using "kubectl port-forward"

Rajpratik71:
## Related Issue

NA

## Proposed Changes

fix logic to access Kubecost UI using "kubectl port-forward"


thomasvn (approved):
Thanks @Rajpratik71 !

PR #1122, opened 2024-08-30
Update info regarding adding productkey for multicluster envs.

thomasvn:
## Proposed Changes

- Add note that in the event you need to view/repair data on a secondary, a productkey may be needed
- Refactor other parts of doc


patrick-tavares (approved):
Looks good to me

PR #1141, opened 2024-10-04
Adds back Hitach and Storj storage configurations

srpomeroy:
## Proposed Changes

Adds back the lost storage configuration when we deprecated the ETL Backup docs from the 2.x branch.
https://github.com/kubecost/docs/blob/ce23be12ba50c6e7b9089956324135f42b5beb7f/install-and-configure/install/etl-backup/etl-backup.md?plain=1#L131-L191


thomasvn (approved):
Thanks for adding back in these storage config options @srpomeroy!

thomasvn:
@srpomeroy Looks like there is one error being thrown by the link checker

chipzoller:
Fixed.

PR #1144, opened 2024-10-10
update documentation to include installation with in-cluster prometheus in openshift

mittal-ishaan:
## Proposed Changes
This PR adds the documentation for installing kubecost in openshift clusters without having the bundled prometheus but using the in-cluster prometheus.

## Depends upon:
https://github.com/opencost/opencost/pull/2944
https://github.com/kubecost/cost-analyzer-helm-chart/pull/3700


install-and-configure/install/provider-installations/openshift-kubecost-install.md
```diff
@@ -60,6 +60,41 @@ If you have not opted to do so during installation, it may be necessary to creat
...
+
+```sh
+oc label namespace kubecost openshift.io/cluster-monitoring=true
+```
+
+2. Install Kubecost with the following command:
+
+```sh
+helm upgrade --install kubecost kubecost/cost-analyzer -n kubecost --create-namespace \
+-f https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/develop/cost-analyzer/values-openshift.yaml \
```
chipzoller, line 77: We need to not tell users to fetch values from the `develop` branch as this will inevitably result in a mismatch at some point and an unsupported configuration. I suggest either a placeholder value or an export to capture the release corresponding to branch name. I know we do this elsewhere in the docs and it's on my to-do list to fix.

chipzoller (changes requested):
Thanks for doing this!

install-and-configure/install/provider-installations/openshift-kubecost-install.md
mittal-ishaan: understood. Makes sense. I should go ahead with adding a placeholder in that case. I see we are doing that in some places

chipzoller:
Are all comments resolved? If so, please resolve.

install-and-configure/install/provider-installations/openshift-kubecost-install.md
```diff
@@ -60,6 +60,26 @@ If you have not opted to do so during installation, it may be necessary to creat
 
 After installation, wait for all pods to be ready. Kubecost will begin collecting data and may take up to 15 minutes for the UI to reflect the resources in the local cluster.
 
+### Using in-cluster Prometheus
+
+Kubecost can be installed without the bundled Prometheus and using the in-cluster Prometheus of OpenShift. This method requires some additional setup steps but allows for better integration with existing OpenShift monitoring infrastructure without the need to create an additional prometheus instance.
```
thomasvn, line 65:
```suggestion
If required Kubecost can leverage an existing Prometheus that exists on your cluster, as opposed to installing Kubecost's bundled Prometheus.
```

install-and-configure/install/provider-installations/openshift-kubecost-install.md
```diff
@@ -42,7 +42,7 @@ Install Kubecost using OpenShift specific values. Note that the below command fe
 
 ```sh
 helm upgrade --install kubecost kubecost/cost-analyzer -n kubecost --create-namespace \
--f https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/develop/cost-analyzer/values-openshift.yaml
+-f https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/<$VERSION>/cost-analyzer/values-openshift.yaml
```
thomasvn, line 45: The drawback to doing this is that now this link will no longer be checked by the CI's link-checker. What happens if we someday decide to rename or delete `values-openshift.yaml`?

Although referencing `develop` branch comes with drawbacks, I still think it's more valuable than putting `<$VERSION> here. @mittal-ishaan @chipzoller thoughts?

install-and-configure/install/provider-installations/openshift-kubecost-install.md
```diff
@@ -60,6 +60,26 @@ If you have not opted to do so during installation, it may be necessary to creat
 
 After installation, wait for all pods to be ready. Kubecost will begin collecting data and may take up to 15 minutes for the UI to reflect the resources in the local cluster.
 
+### Using in-cluster Prometheus
+
```
thomasvn, line 64: We generally do not recommend users leverage an existing Prometheus because this often leads to numerous downstream issues. I think it would be helpful to add the following warning in this section.

```suggestion

{% hint style="warning" %} 
This installation method is available, but not generally recommended. Please review the following documentation before proceeding. [Documentation](/install-and-configure/advanced-configuration/custom-prom).
{% endhint %}

```
```

File: install-and-configure/install/provider-installations/openshift-kubecost-install.md (hunk 3 above)
Comment (mittal-ishaan) [reply install-and-configure/install/provider-installations/openshift-kubecost-install.md:45]: We can develop. But we can also start with having versions. like `v2.5` and then update docs every new minor or major version gets released.

Comment (chipzoller) [reply install-and-configure/install/provider-installations/openshift-kubecost-install.md:45]: @thomasvn, the problem here with listing files from the `develop` branch is we put users in a precarious and potentially unsupported position when they install a versioned release but use a values file which doesn't correspond to that release. We really need to get away from that practice entirely when we're talking about our official docs. Rather than putting in a placeholder like `<$VERSION>` we should make this the same as the version referenced by the docs. For example, if users are reading version 2.5 of our docs then any files that are associated with that release should be pointed to the `v2.5` branch. This obviously means that any new files that get referenced are "released" in that branch.

Comment (thomasvn) [reply install-and-configure/install/provider-installations/openshift-kubecost-install.md:45]: @chipzoller Ok good point. So @mittal-ishaan it should probably be something like the following?

```suggestion
-f https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/v2.5/cost-analyzer/values-openshift.yaml
```

thomasvn: @mittal-ishaan ?

mittal-ishaan: sure, let me update that :+1:

thomasvn (approved):

PR #1145, opened 2024-10-14
updated screenshot to show net amortized in Aggregate cost

KarolTegha:
Updated screenshot to show that we are using Net Amortized costs.


srpomeroy (commented):
We may need to be careful here. In the past we've experienced situations where customers don't have net costs in their CUR. I am not sure if the experience in Cost Explorer changes in those situations.

[See this doc](https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer/cloud-cost-metrics#aws-cost-metrics) for additional details in how we consume cloud costs from the CUR.

chipzoller:
Need a decision whether to merge or close by 2/14.

PR #1148, opened 2024-10-24
[2.5] Shared GPU support and GPU Optimization

chipzoller:
Signed-off-by: Chip Zoller <chipzoller@gmail.com>

## Proposed Changes

Initial section on NVIDIA shared GPU support and the GPU savings insights card.


install-and-configure/advanced-configuration/gpu.md
```diff
@@ -308,3 +308,34 @@ kubectl -n kubecost port-forward svc/kubecost-prometheus-server 8080:80
...
 
 ![Prometheus query showing DCGM Exporter metric](/images/gpu-prometheus-query.png)
+
+## Shared GPU Support
+
+Kubecost supports NVIDIA GPU sharing using either the CUDA [time-slicing](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-sharing.html) or [Multi-Process Service (MPS)](https://docs.nvidia.com/deploy/mps/index.html) methods. MIG is currently unsupported but is being evaluated for a future release. When employing either time-slicing or MPS, you must use the `renameByDefault=true` option in the [NVIDIA device plugin's](https://github.com/NVIDIA/k8s-device-plugin?tab=readme-ov-file#shared-access-to-gpus) configuration stanza. This parameter instructs the device plugin to advertise the resource `nvidia.com/gpu.shared` on nodes where GPU sharing is enabled. Without this configuration option, the device plugin will instead advertise `nvidia.com/gpu` which will mean Kubecost is unable to disambiguate an "exclusive" GPU access request from a shared GPU access request. As a result, Kubecost's cost information will be inaccurate.
+
+The following is an example of a time-slicing configuration which sets the `renameByDefault` parameter.
```
thomasvn, line 316:
```suggestion
The following is an example of a time-slicing configuration which sets the `renameByDefault` parameter in the NVIDIA device plugin helm chart.
```

Not sure if the wording update is good, but I think it could be helpful to clarify what the below YAML block is configuring.

install-and-configure/advanced-configuration/gpu.md
```diff
@@ -308,3 +308,34 @@ kubectl -n kubecost port-forward svc/kubecost-prometheus-server 8080:80
...
+$ kubectl describe node mynodename
+...
+Capacity:
+  nvidia.com/gpu.shared: 16
+...
+```
+
+### Limitations
+
+There are limitations of which to be aware when using NVIDIA GPU sharing with either time-slicing or MPS. Because [NVIDIA does not support](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-sharing.html#limitations) providing utilization metrics via DCGM Exporter for containers using shared GPUs, Kubecost will display a GPU cost of zero for these workloads. However, the [GPU Savings Optimization](/using-kubecost/navigating-the-kubecost-ui/savings/gpu-optimization.md) card (Kubecost Enterprise) will be able to indicate in the utilization table which containers are configured for GPU sharing providing some visibility.
```
thomasvn, line 341: Is it important to put these limitations at the top of the "Shared GPU" section? Before users begin making configuration changes to the NVIDIA device plugin.

chipzoller: Good point, although I think an info box or sentence linking to the Limitations section here may be all we need.

chipzoller: Added a warning note. See if that works.

kwombach12 (approved):
these look really good to me

install-and-configure/advanced-configuration/gpu.md
thomasvn: Oh ok got it! Should we specify that somewhere? It's not immediately clear to me how to use the YAML from this paragraph alone. I had to dig through the `k8s-device-plugin` repo to better understand what this was for.

install-and-configure/advanced-configuration/gpu.md
thomasvn: This works great, thank you!

thomasvn (approved):
Generally this looks great to me! I only had a minor comment about the code block clarifying `renameByDefault=true`.

install-and-configure/advanced-configuration/gpu.md
chipzoller: I asked myself the same question but opted not to turn this into a full "let's teach you everything about NVIDIA GPU sharing soup to nuts" section because 1) it's a complicated subject and 2) it would be very lengthy. I think I'm ok, at least for now, leaving it with just the essentials and seeing where we go from here.

thomasvn: Works for me!

PR #1156, opened 2024-12-02
Add GPU note on Efficiency dashboard

mmurph3:
## Proposed Changes

- Makes minor changes to the GPU Savings Insights page.
- Adds a note to the Efficiency dashboard page to help troubleshoot the missing GPU column.


chipzoller (commented):
First sentence isn't exactly true. Second one belongs on the Efficiency page.

thomasvn:
Thanks @mmurph3 for starting this! This is an effort to address some confusion raised by users in [SUP-6416](https://kubecost.atlassian.net/browse/SUP-6416).

Agree with Chip here that the first bullet point isn't necessarily true. It's probably safe to remove it.

The second bullet point is good. How about we modify it slightly to the following?

```md
## Troubleshooting

### Kubecost dashboards not showing GPU Efficiency or GPU Savings

In order for Kubecost to begin displaying GPU features, it must first detect that **at least one** of your clusters has a nonzero amount of GPU usage. Please validate that DCGM-Exporter is running in the clusters which have GPUs and that Kubecost is scraping nonzero GPU metrics from the exporter.
```

It may be good to add this "Troubleshooting" section to this doc here, as well as the Efficiency doc we have. https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/efficiency-dashboard

chipzoller:
If we're going to create a public Troubleshooting section specific to GPU, we may want to take this opportunity to build it out more completely à la what I have put together [here](https://app.gitbook.com/o/MQuX6uFwV0j7vIHtR15E/s/xLM07kCOoiNtRubOhU77/customer-nvidia-gpu-troubleshooting#no-gpu-column-in-efficiency-page) (internal resource).

mmurph3:
Made some changes before seeing the most recent comments. If we don't agree with what I wrote, I'm ok with changing/moving it. I agree, a built out troubleshooting doc would be good.

@chipzoller , for some reason I'm getting 403'd on that internal link you gave: https://app.gitbook.com/o/MQuX6uFwV0j7vIHtR15E/s/xLM07kCOoiNtRubOhU77/customer-nvidia-gpu-troubleshooting#no-gpu-column-in-efficiency-page

I can see about getting access through Cliff.

thomasvn:
@mmurph3 These are good changes, but are you sure it's enough? I'm concerned these small add-ons may be missed by some users. Hard to catch the sentence in a long document.

What do you think about adding a "Troubleshooting" section to both these docs, and filling in a bit more details about what to do in the event that they are not seeing GPU Efficiency Features?

chipzoller:
If you guys don't mind, I'd like to take this over and propose some changes here. It just will have to be next week.

thomasvn:
@chipzoller Good with me!

chipzoller:
Requesting review from @thomasvn and @mmurph3.

mmurph3 (commented):
lgtm

using-kubecost/navigating-the-kubecost-ui/savings/gpu-optimization.md
```diff
@@ -32,13 +32,13 @@ Clicking on each recommendation tile displays a window with further details on t
...
+In the first version of the GPU Optimization Savings Insights card there are a few known limitations.
 
 - Multiple containers with the same name and running on the same cluster, node, and namespace combination (i.e., "identical" containers) might result in the following effects:
   - The savings number provided on Optimize and Remove cards may be an implicit sum of the total cost these containers.
   - Recommendations will only be provided for one of them.
   - The utilization table may not show these identical containers.
 - GPU nodes must be running or have run at least one container utilizing a GPU for it to be represented on the utilization table in either the Cluster aggregation’s GPU nodes column or on the Node aggregation.
-- Optimize may be as accurate as possible in certain cases since Kubecost currently infers utilization about all GPUs from a single averaged utilization number.
+- The Optimize recommendation may not be as accurate as possible in certain cases since Kubecost currently infers utilization about all GPUs from a single averaged utilization number.
```
thomasvn, line 42:
```suggestion
- The Optimize recommendation may not be as accurate as possible in certain cases since Kubecost currently infers utilization about each GPU from its averaged utilization number.
```

Would this slight change be more accurate? For me, it adds clarity to the sentence.

chipzoller: To me this suggests Kubecost is looking at each GPU and then determining its average utilization which isn't what's happening here. It's getting one number back which is an implicit average of all GPUs, which is what we need to improve for the next version so that it _can_ look at each GPU individually.

thomasvn: Ok I see, makes sense! Let's leave the wording as is.

thomasvn (approved):

PR #1161, opened 2024-12-11
Deletes API doc describing cost-model cache

thomasvn:
## Related Issue

- These APIs will be removed from implementation by v2.6.0 https://github.com/opencost/opencost/pull/3016
- Doc first introduced by

## Proposed Changes

- Remove doc from `SUMMARY.md`
- Should we also delete the entire doc from the repo? @chipzoller @kwombach12 @srpomeroy Not sure what our standard should be here. I'm hesitant to do so because this docs repo does not currently do release branches (e.g. `v2.4`, `v2.5`)


srpomeroy:
@thomasvn re: 
> Should we also delete the entire doc from the repo? @chipzoller @kwombach12 @srpomeroy Not sure what our standard should be here. I'm hesitant to do so because this docs repo does not currently do release branches (e.g. v2.4, v2.5)

I don't have a strong opinion one way or another. Just removing from summary.md seems safe for now so we can retain the doc here in GH. But our docs platform is changing soon. I'm not sure the new solution will have the same luxury. I am also unsure how that change will impact our want/need to do versioned docs.

thomasvn:
Thanks @srpomeroy! Yeah let's just remove from `SUMMARY.md` for now. I definitely _hope_ we'll be able to set up versioned docs soon. Let's wait until we migrate to the new docs solution.

chipzoller:
I think it's safe to remove entirely. When we move to versioned docs on IBM's side, we would do that going forward and not retroactively. Their system also doesn't use git so it's an entirely different beast.

thomasvn:
Ok understood, thanks @chipzoller !

chipzoller:
Good to merge this for 2.6?

thomasvn:
@chipzoller Good to merge! Just needs approval.

PR #1172, opened 2025-01-27
Ignore jwt.io on link check

thomasvn:
## Related Issue

N/A

## Proposed Changes

This link caused failing CI checks. It's likely that https://jwt.io began blocking non-human requests with HTTP 500s.


thomasvn:
Link checker has passed!

PR #1176, opened 2025-02-06
Remove default for kubecostAggregator.dbReadThreads` thread limit on read database

biancaburtoiu:
## Related Issue
https://github.com/kubecost/cost-analyzer-helm-chart/pull/3845

## Proposed Changes

The helm chart PR above modifies the `kubecostAggregator.dbReadThreads` default value. This PR updates the public documentation to reflect this change.

### DIFF
--- a/install-and-configure/advanced-configuration/resource-consumption.md
+++ b/install-and-configure/advanced-configuration/resource-consumption.md
@@ -81,6 +81,8 @@ kubecostAggregator:
   dbConcurrentIngestionCount: 1
 ```
 
+Note that by default, `dbReadThreads` is set to `0` which means that the number of threads that can be used by the read database is bound by the number of cores on the host node.
+
 **Set rough memory limits.** By default these are set to `0GB` which means no limit. Once a baseline memory usage is established, it can be helpful to set these limits such that `dbMemoryLimit + dbWriteMemoryLimit <= memoryAvailableToAggregatorPod`.
 
 {% hint style="warning" %}
--- a/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
+++ b/install-and-configure/install/multi-cluster/federated-etl/aggregator.md
@@ -83,10 +83,10 @@ kubecostAggregator:
   etlDailyStoreDurationDays: 91
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If increasing this value, it is recommended to increase the
-  # aggregator's memory requests & limits.
-  # default: 1
-  dbReadThreads: 1
+  # UI queries). If value is 0, the number of threads is bound by the
+  # number of cores
+  # default: 0 is no limit
+  dbReadThreads: 0
 
   # How many threads the write database is configured with (i.e. ingestion of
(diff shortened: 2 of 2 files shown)


install-and-configure/advanced-configuration/resource-consumption.md
```diff
@@ -81,6 +81,8 @@ kubecostAggregator:
   dbConcurrentIngestionCount: 1
```
biancaburtoiu, line 81: We are recommending using `dbReadThreads: 1` above. I don't think it's necessary to change this recommendation but open to discussion

install-and-configure/install/multi-cluster/federated-etl/aggregator.md
```diff
@@ -83,10 +83,10 @@ kubecostAggregator:
   etlDailyStoreDurationDays: 91
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If increasing this value, it is recommended to increase the
-  # aggregator's memory requests & limits.
-  # default: 1
-  dbReadThreads: 1
+  # UI queries). If value is 0, the number of threads is bound by the
```
ameijer, line 86:
```suggestion
  # UI queries). If value is 0, the number of threads is bounded by the
```

install-and-configure/advanced-configuration/resource-consumption.md
```diff
@@ -81,6 +81,8 @@ kubecostAggregator:
...
 ```
 
+Note that by default, `dbReadThreads` is set to `0` which means that the number of threads that can be used by the read database is bound by the number of cores on the host node.
```
ameijer, line 84:
```suggestion
Note that by default, `dbReadThreads` is set to `0` which means that the number of threads that can be used by the read database is bounded by the number of cores on the host node.
```

ameijer (dismissed):

install-and-configure/advanced-configuration/resource-consumption.md
thomasvn: There's some text above "By default these should all be set to `1`". We should delete this since it's no longer true?

biancaburtoiu: You're right, thanks for calling that out. The other two values still have those defaults, though. 
WDYT if we rephrase from 
"**Lower concurrency.** By default these should all be set to `1`."
to
"Settings for lower concurrency:" 
or something like that?

thomasvn: What about this? To remain consistent with how it's worded in all the other paragraphs above?

```
**Lower concurrency.** Set values to `1`.
```

biancaburtoiu: Applied, thank you!

thomasvn (approved):
Looks good to me. Thanks @biancaburtoiu!

chipzoller:
I have incorporated the changes defined in this PR in the docs and is reflected on the test site [here](https://ibmdocs-test.dcs.ibm.com/docs/en/kubecost/self-hosted/2.x). These changes will get pushed to production upon release of 2.7.0. Closing.

PR #1185, opened 2025-05-28
Removal of Assets Diff API

thomasvn:
## Related Issue

N/A

## Proposed Changes

This API has been removed since v2.0 https://github.com/kubecost/cost-analyzer-helm-chart/blob/v2.0/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L409-L417


HMetcalfeW (approved):
Now we need to roll this up to Adobe Experience Manager so the public can see it. Thank you for doing this, @thomasvn :)

PR #1194, opened 2025-10-23
removed docs

nandrews-ibm:
As discussed with docs team, with [docs.kubecost.com](docs.kubecost.com) being the source of truth going forward, it might make sense to deprecate this repo to not have to keep two sets of documentation in sync and simply use this repo as a pointer for folks who still would navigate here.


HMetcalfeW (changes requested):
@nandrews-ibm , let's get a tag of the repo before we delete anything. Perhaps we should update the ReadMe to point users to the right instruction resources in Confluence?

thomasvn:
Thanks @nandrews-ibm for this cleanup. According to discussion on Slack, we are ok archiving this repository, because AEM is now the source of truth as opposed to Github.

Thanks for the README change! I've also tagged the latest commit in this repo with tag `final`. If users still want to browse the old content on this repo, they can browse on the tag `final`. Example: https://github.com/kubecost/docs/blob/final/apis/monitoring-apis/api-allocation.md

@HMetcalfeW @nandrews-ibm Let me know what you think about this approach!

HMetcalfeW (approved):
LGTM

